PR #6, opened 2019-11-19
Fix GCC 8 build

mbautin:
We were not actually using the GCC 8 compiler, but were using the default GCC 5.5.
Also remove the dependency on s3cmd because we are not downloading dependencies from S3.


mbautin:
We were not actually using the GCC 8 compiler, but were using the default GCC 5.5.
Also remove the dependency on s3cmd because we are not downloading dependencies from S3.

PR #10, opened 2020-04-21
[#3935] Bump up cqlsh version

iSignal:
Bump up to latest cqlsh version to pick up commit https://github.com/yugabyte/cqlsh/commit/6374f27d6793b13a3e1551b4a7426b7e7849e7c5


jaki:
I'm assuming you spoke with @mbautin on updating this rather than <https://github.com/yugabyte/yugabyte-db/blob/master/thirdparty/build_definitions/cqlsh.py>.  Would be interested to know more details on that here if possible.

iSignal:
I assumed that both places need to be updated (though, tbh, I'm not sure why). The other part in the main repo is in a diff that you are subscribed to.

PR #31, opened 2020-06-18
Sanity check dynamic shared library dependencies

svarnau:
yugabyte/yugabyte-db#4774

* Add library scan at the end of the build script.
* Fix up icu executables on mac.
* Fix up redis library on mac.

### DIFF
--- a/yb_build_thirdparty_main.py
+++ b/yb_build_thirdparty_main.py
@@ -958,6 +958,87 @@ class Builder:
         ]
         return openssl_options
 
+class LibTest:
+    """
+    Verify correct library paths are used in installed dynamically-linked eecutables and
+    libraries.
+    """
+    def __init__(self):
+        self.tp_dir = os.path.dirname(os.path.realpath(sys.argv[0]))
+        self.tp_installed_dir = os.path.join(self.tp_dir, 'installed')
+        if is_mac():
+            self.tool = "otool -L"
+            self.okay_paths = "^\t("
+            self.okay_paths += "/usr/|"
+            self.okay_paths += "/System/Library/|"
+            self.okay_paths += "@rpath|"
+            self.okay_paths += "@loader_path|"
+            self.okay_paths += f"{self.tp_dir})"
+        else:
+            self.tool = "ldd"
+            self.okay_paths = "^\tlinux-vdso|"
+            self.okay_paths += "^\t/lib64/|"
+            self.okay_paths += "^\tstatically linked|"
+            self.okay_paths += "^.* => /lib64/|"
+            self.okay_paths += "^.* => /opt/yb-build/brew/linuxbrew|"
+            self.okay_paths += f"^.* => {self.tp_dir}"
+
+    def good_libs(self,file_path):
+        status = True
+        if is_mac():
+            libout = subprocess.check_output(['otool', '-L' ] + [file_path]).decode('utf-8')
+            if 'is not an object file' in libout:
+                return status
+        else:
+            # SSL libs need to search in common lib
+            ld_search = {"LD_LIBRARY_PATH": self.tp_installed_dir + "/common/lib" }
+            try:
+                libout = subprocess.check_output(['ldd'] + [file_path],
+                                                 env=ld_search,
+                                                 stderr=subprocess.DEVNULL).decode('utf-8')
+            except subprocess.CalledProcessError as cpe:
+                # Normal error code if file is not a shared library or executable
+                if cpe.returncode != 1:
+                    print("Unexpected error for file:",file_path)
+                    status = False
+                return status
+
+        for line in libout.splitlines():
+            # Dependent libs indented with tab
+            if line.startswith('\t'):
+                if not re.match(self.okay_paths,line):
+                    if status:
+                        print(file_path + ":")
+                    print("Bad path:",line)
+                    status = False
+        return status
+
+    def run(self):
+        print("Scanning installed executables and libraries...")
+        test_pass = True
+        # files to examine are much reduced if we look only at bin and lib directories
+        dir_pattern = '^(lib|libcxx|[s]bin)$'
+        dirs = [os.path.join(self.tp_installed_dir, type) for type in BUILD_TYPES]
+        for installed_dir in dirs:
+            with os.scandir(installed_dir) as candidate_dirs:
+                for candidate in candidate_dirs:
+                    if re.match(dir_pattern, candidate.name):
+                        examine_path = os.path.join(installed_dir,candidate.name)
+                        for dirpath, dirnames, files in os.walk(examine_path):
+                            for file_name in files:
+                                full_path = os.path.join(dirpath,file_name)
+                                if os.path.islink(full_path):
+                                    continue
+                                if not self.good_libs(full_path):
+                                    test_pass = False
+        if not test_pass:
+            print('-------------------')
+            print("Found problematic library dependencies, using tool: ",self.tool)
+            print('-------------------')
+        else:
+            print("No problems found with library dependencies.")
+        return test_pass
+
 
 def main():
     unset_if_set('CC')
@@ -972,6 +1053,10 @@ def main():
     builder = Builder()
     builder.init()
     builder.run()
+
+    tester = LibTest()
+    if not tester.run():
+        sys.exit(1)
 
 
 if __name__ == "__main__":
--- a/yb_build_thirdparty_main.py
+++ b/yb_build_thirdparty_main.py
@@ -967,16 +967,17 @@ class LibTestBase:
         self.tp_dir = os.path.dirname(os.path.realpath(sys.argv[0]))
         self.tp_installed_dir = os.path.join(self.tp_dir, 'installed')
 
+    def init_regex(self):
+        self.okay_paths = re.compile("|".join(self.lib_re_list))
+
     def check_lib_deps(self, file_path, cmdout):
         status = True
         for line in cmdout.splitlines():
-            # Dependent libs indented with tab
-            if line.startswith('\t'):
-                if not re.match(self.okay_paths,line):
-                    if status:
-                        log(file_path + ":")
-                    log("Bad path: %s", line)
-                    status = False
+            if not self.okay_paths.match(line):
+                if status:
+                    log(file_path + ":")
+                log("Bad path: %s", line)
+                status = False
         return status
 
     # overridden in platform specific classes
@@ -984,15 +985,16 @@ class LibTestBase:
         pass
 
     def run(self):
-        log("Scanning installed executables and libraries...")
+        self.init_regex()
+        heading("Scanning installed executables and libraries...")
         test_pass = True
         # files to examine are much reduced if we look only at bin and lib directories
-        dir_pattern = '^(lib|libcxx|[s]bin)$'
+        dir_pattern = re.compile('^(lib|libcxx|[s]bin)$')
         dirs = [os.path.join(self.tp_installed_dir, type) for type in BUILD_TYPES]
         for installed_dir in dirs:
             with os.scandir(installed_dir) as candidate_dirs:
                 for candidate in candidate_dirs:
-                    if re.match(dir_pattern, candidate.name):
+                    if dir_pattern.match(candidate.name):
                         examine_path = os.path.join(installed_dir, candidate.name)
                         for dirpath, dirnames, files in os.walk(examine_path):
                             for file_name in files:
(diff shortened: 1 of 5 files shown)


yb_build_thirdparty_main.py
```diff
@@ -958,6 +958,87 @@ def get_openssl_related_cmake_args(self):
         ]
         return openssl_options
 
+class LibTest:
+    """
+    Verify correct library paths are used in installed dynamically-linked eecutables and
```
mbautin, line 963: typo: executables

yb_build_thirdparty_main.py
```diff
@@ -958,6 +958,87 @@ def get_openssl_related_cmake_args(self):
...
+        self.tp_dir = os.path.dirname(os.path.realpath(sys.argv[0]))
+        self.tp_installed_dir = os.path.join(self.tp_dir, 'installed')
+        if is_mac():
+            self.tool = "otool -L"
+            self.okay_paths = "^\t("
+            self.okay_paths += "/usr/|"
+            self.okay_paths += "/System/Library/|"
+            self.okay_paths += "@rpath|"
+            self.okay_paths += "@loader_path|"
+            self.okay_paths += f"{self.tp_dir})"
```
mbautin, line 976: It would probably be better to assign these "okay paths" to a list and then join it with `|` once outside of the mac vs. linux `if` statement.

yb_build_thirdparty_main.py
```diff
@@ -958,6 +958,87 @@ def get_openssl_related_cmake_args(self):
...
+        else:
+            self.tool = "ldd"
+            self.okay_paths = "^\tlinux-vdso|"
+            self.okay_paths += "^\t/lib64/|"
+            self.okay_paths += "^\tstatically linked|"
+            self.okay_paths += "^.* => /lib64/|"
+            self.okay_paths += "^.* => /opt/yb-build/brew/linuxbrew|"
+            self.okay_paths += f"^.* => {self.tp_dir}"
+
+    def good_libs(self,file_path):
```
mbautin, line 986: Nit: space after `,`

yb_build_thirdparty_main.py
```diff
@@ -958,6 +958,87 @@ def get_openssl_related_cmake_args(self):
...
+            # SSL libs need to search in common lib
+            ld_search = {"LD_LIBRARY_PATH": self.tp_installed_dir + "/common/lib" }
+            try:
+                libout = subprocess.check_output(['ldd'] + [file_path],
+                                                 env=ld_search,
+                                                 stderr=subprocess.DEVNULL).decode('utf-8')
+            except subprocess.CalledProcessError as cpe:
+                # Normal error code if file is not a shared library or executable
+                if cpe.returncode != 1:
+                    print("Unexpected error for file:",file_path)
```
mbautin, line 1002: Nit: add a space after `,`

yb_build_thirdparty_main.py
```diff
@@ -958,6 +958,87 @@ def get_openssl_related_cmake_args(self):
...
+
+    def good_libs(self,file_path):
+        status = True
+        if is_mac():
+            libout = subprocess.check_output(['otool', '-L' ] + [file_path]).decode('utf-8')
+            if 'is not an object file' in libout:
+                return status
+        else:
+            # SSL libs need to search in common lib
+            ld_search = {"LD_LIBRARY_PATH": self.tp_installed_dir + "/common/lib" }
```
mbautin, line 994: We should not have to set LD_LIBRARY_PATH, we should be configuring rpaths correctly. How did this come up?

yb_build_thirdparty_main.py
```diff
@@ -958,6 +958,87 @@ def get_openssl_related_cmake_args(self):
...
+            try:
+                libout = subprocess.check_output(['ldd'] + [file_path],
+                                                 env=ld_search,
+                                                 stderr=subprocess.DEVNULL).decode('utf-8')
```
mbautin, line 998: Putting stderr to /dev/null often hides useful diagnostic messages.

yb_build_thirdparty_main.py
```diff
@@ -958,6 +958,87 @@ def get_openssl_related_cmake_args(self):
...
+        status = True
+        if is_mac():
+            libout = subprocess.check_output(['otool', '-L' ] + [file_path]).decode('utf-8')
+            if 'is not an object file' in libout:
+                return status
+        else:
+            # SSL libs need to search in common lib
+            ld_search = {"LD_LIBRARY_PATH": self.tp_installed_dir + "/common/lib" }
+            try:
+                libout = subprocess.check_output(['ldd'] + [file_path],
```
mbautin, line 996: `['ldd', file_path]`

yb_build_thirdparty_main.py
```diff
@@ -958,6 +958,87 @@ def get_openssl_related_cmake_args(self):
         ]
         return openssl_options
 
+class LibTest:
```
mbautin, line 961: I think this class is better split into a hierarchy of a base class and Linux / macOS specific checkers.

yb_build_thirdparty_main.py
```diff
@@ -958,6 +958,87 @@ def get_openssl_related_cmake_args(self):
...
+                            for file_name in files:
+                                full_path = os.path.join(dirpath,file_name)
+                                if os.path.islink(full_path):
+                                    continue
+                                if not self.good_libs(full_path):
+                                    test_pass = False
+        if not test_pass:
+            print('-------------------')
+            print("Found problematic library dependencies, using tool: ",self.tool)
+            print('-------------------')
```
mbautin, line 1037: Report errors using the `log` function as done elsewhere in this codebase.

yb_build_thirdparty_main.py
```diff
@@ -958,6 +958,87 @@ def get_openssl_related_cmake_args(self):
...
+        else:
+            print("No problems found with library dependencies.")
```
mbautin, line 1039: ditto

yb_build_thirdparty_main.py
```diff
@@ -958,6 +958,87 @@ def get_openssl_related_cmake_args(self):
...
+        print("Scanning installed executables and libraries...")
+        test_pass = True
+        # files to examine are much reduced if we look only at bin and lib directories
+        dir_pattern = '^(lib|libcxx|[s]bin)$'
+        dirs = [os.path.join(self.tp_installed_dir, type) for type in BUILD_TYPES]
+        for installed_dir in dirs:
+            with os.scandir(installed_dir) as candidate_dirs:
+                for candidate in candidate_dirs:
+                    if re.match(dir_pattern, candidate.name):
+                        examine_path = os.path.join(installed_dir,candidate.name)
```
mbautin, line 1026: space after comma

yb_build_thirdparty_main.py
```diff
@@ -958,6 +958,87 @@ def get_openssl_related_cmake_args(self):
...
+                        for dirpath, dirnames, files in os.walk(examine_path):
+                            for file_name in files:
+                                full_path = os.path.join(dirpath,file_name)
```
mbautin, line 1029: space after comma

yb_build_thirdparty_main.py
svarnau: True, but in this case, the only message it seems to provide is a warning that a file is not executable, which is usually on non-dynamically-linked files, which we just want to ignore.  If a file is corrupted such that ldd has problems interpreting it, then I think there are other build issues beyond scope of this check.

yb_build_thirdparty_main.py
```diff
@@ -958,6 +958,99 @@ def get_openssl_related_cmake_args(self):
...
+                       "^.* => /lib64/",
+                       "^.* => /opt/yb-build/brew/linuxbrew",
+                       f"^.* => {self.tp_dir}"]
+        self.okay_paths = "|".join(path_regex)
+
+    def good_libs(self, file_path):
+        try:
+            # Hide messages about execute permissions
+            libout = subprocess.check_output(['ldd', file_path],
+                                             stderr=subprocess.DEVNULL).decode('utf-8')
```
mbautin, line 1045: Let's not suppress stderr, even if it prints a lot of harmless messages. We might miss important error messages in case of a failure.

Or, alternatively, capture both stdout and stderr, e.g. using 
```python
    proc = subprocess.Popen(cmd,
        stdout = subprocess.PIPE,
        stderr = subprocess.PIPE,
    )
    stdout, stderr = proc.communicate()
```
and print stderr to `sys.stderr` only if it does not contain the expected harmless message.

yb_build_thirdparty_main.py
```diff
@@ -958,6 +958,99 @@ def get_openssl_related_cmake_args(self):
...
+        except subprocess.CalledProcessError as cpe:
+            # Normal error code if file is not a shared library or executable
+            if cpe.returncode != 1:
+                log("Unexpected error for: ldd", file_path)
```
mbautin, line 1049: Include the error code in the log message. Also I think file_path won't be logged properly this way.
Has to be something like:
```python
log("Unexpected exit code %d from ldd, file %s", cpe.returncode, file_path)
```

yb_build_thirdparty_main.py
```diff
@@ -958,6 +958,99 @@ def get_openssl_related_cmake_args(self):
...
+
+    # overridden in platform specific classes
+    def good_libs(self, file_path):
+        pass
+
+    def run(self):
+        log("Scanning installed executables and libraries...")
+        test_pass = True
+        # files to examine are much reduced if we look only at bin and lib directories
+        dir_pattern = '^(lib|libcxx|[s]bin)$'
```
mbautin, line 990: Compile the regex once using `re.compile`

yb_build_thirdparty_main.py
```diff
@@ -958,6 +958,99 @@ def get_openssl_related_cmake_args(self):
...
+    def __init__(self):
+        self.tp_dir = os.path.dirname(os.path.realpath(sys.argv[0]))
+        self.tp_installed_dir = os.path.join(self.tp_dir, 'installed')
+
+    def check_lib_deps(self, file_path, cmdout):
+        status = True
+        for line in cmdout.splitlines():
+            # Dependent libs indented with tab
+            if line.startswith('\t'):
+                if not re.match(self.okay_paths,line):
```
mbautin, line 975: Compile the okay_paths regex once using `re.compile`

yb_build_thirdparty_main.py
```diff
@@ -973,6 +1066,12 @@ def main():
     builder.init()
     builder.run()
 
+    if is_mac():
+        tester = LibTestMac()
+    else:
```
mbautin, line 1071:
```python
if is_mac():
    tester = LibTestMac()
elif is_linux():
    tester = LibTestLinux()
else:
   fatal(f"Unsupported platform: {format(platform.system()}")
```

yb_build_thirdparty_main.py
```diff
@@ -958,6 +958,99 @@ def get_openssl_related_cmake_args(self):
...
+    def __init__(self):
+        super().__init__()
+        self.tool = "ldd"
+        path_regex = ["^\tlinux-vdso",
+                       "^\t/lib64/",
+                       "^\t/opt/yb-build/brew/linuxbrew",
+                       "^\tstatically linked",
+                       "^.* => /lib64/",
+                       "^.* => /opt/yb-build/brew/linuxbrew",
+                       f"^.* => {self.tp_dir}"]
```
mbautin, line 1038: It would probably be better to assign this list itself to a field like `self.okay_path_patterns` in the derived class constructor, and then join it with `|` and compile the regex once in the base class's `run()` method.

yb_build_thirdparty_main.py
```diff
@@ -958,6 +958,99 @@ def get_openssl_related_cmake_args(self):
...
+        self.okay_paths = "|".join(path_regex)
```
mbautin, line 1039: Then these lines wouldn't have to be repeated in the two subclasses either.

yb_build_thirdparty_main.py
```diff
@@ -958,6 +958,100 @@ def get_openssl_related_cmake_args(self):
...
+                            f"^.* => {self.tp_dir}"]
+
+    def good_libs(self, file_path):
+        proc = subprocess.run(['ldd', file_path], capture_output=True)
+        if proc.returncode > 1:
+            log("Unexpected exit code %d from ldd, file %s", proc.returncode, file_path)
+            log(proc.stdout.decode('utf-8'))
+            log(proc.stderr.decode('utf-8'))
+            return False
+        libout = proc.stdout.decode('utf-8') + proc.stderr.decode('utf-8')
```
mbautin, line 1052: Perhaps add a separator (a newline?) between stdout and stderr.

PR #32, opened 2020-06-25
thirparty_build compatiblity with python 3.6

svarnau:
[yugabyte/yugabyte-db#4872] Fix python compatibility.

Also add library check patterns for Ubuntu.


mbautin:
@svarnau Thanks! It would be good to also add an Ubuntu build on CircleCI.

PR #33, opened 2020-06-25
Add build job for Ubuntu and fix build issues in libcrypt_blowfish

svarnau:
Addresses issues:
- yugabyte/yugabyte-db#4891 (libcrypt_blowfish build issues -- not honoring CFLAGS and CC env vars)
- yugabyte/yugabyte-db#4959 (Ubuntu job for yugabyte-db-thirdparty)


mbautin:
Instead of duplicating the script, we should use a build matrix. https://circleci.com/blog/circleci-matrix-jobs/

patches/crypt_blowfish-makefile-cflags.patch
```diff
@@ -0,0 +1,11 @@
+--- Makefile	2017-08-10 14:39:36.000000000 -0700
++++ Makefile	2020-06-26 11:32:53.000000000 -0700
+@@ -18,7 +18,7 @@
+ AS = $(CC)
+ LD = $(CC)
+ RM = rm -f
+-CFLAGS = -fPIC -W -Wall -Wbad-function-cast -Wcast-align -Wcast-qual -Wmissing-prototypes -Wstrict-prototypes -Wshadow -Wundef -Wpointer-arith -O2 -fomit-frame-pointer -funroll-loops
++CFLAGS = $(CFLAG_EXTRA) -fPIC -W -Wall -Wbad-function-cast -Wcast-align -Wcast-qual -Wmissing-prototypes -Wstrict-prototypes -Wshadow -Wundef -Wpointer-arith -O2 -fomit-frame-pointer -funroll-loops
```
mbautin, line 8: I think EXTRA_CFLAGS is a more accepted name for this kind of an environment variable.

build_and_release.sh
```diff
@@ -47,12 +47,17 @@ echo "Bash version: $BASH_VERSION"
 
 if "$is_mac"; then
   ( set -x; shasum --version )
-else
+elif "$is_centos"; then
   (
     set -x
     sha256sum --version
     libtool --version
```
mbautin, line 54: How are we getting away without libtool on Ubuntu?

build_definitions/crypt_blowfish.py
```diff
@@ -27,11 +27,16 @@ def __init__(self):
...
         self.copy_sources = True
+        self.patches = ['crypt_blowfish-makefile-cflags.patch']
+        self.patch_strip = 0
 
     def build(self, builder):
         log_prefix = builder.log_prefix(self)
         log_output(log_prefix, ['make', 'clean'])
-        log_output(log_prefix, ['make'])
+        if is_mac():
+            log_output(log_prefix, ['make', 'CFLAG_EXTRA="-mmacosx-version-min=10.14"'])
```
mbautin, line 37: I had an impression that we already had this kind of a flag somewhere. Can we deduplicate it so that we can specify the target macOS version once for the entire project?
Also if crypt_blowfish ignores the mechanism that we use to pass flags to other dependencies, could that cause problems? Perhaps we should add other flags here via the same mechanism?

.circleci/config.yml
```diff
@@ -1,6 +1,14 @@
...
+
 jobs:
   build:
+    parameters:
+      docker_image:
+        type: string
+
+    parallelism: 2
+
+    # Need docker run with SYS_PTRACE for gcc/Asan address sanitizer to work
```
mbautin, line 11: This is confusing. We don't run with Address Sanitizer in gcc, we use clang for that purpose.

build_and_release.sh
svarnau: At least when I ran with this docker image, it failed to find libtool pre-installed. It seems most of the tools using configure bring their own libtool.  Running find _build -name libtool_ brings up quite a few of them.

build_definitions/crypt_blowfish.py
svarnau: Good point. That flag, and all of the other standard flags are passed as CXXFLAGS. If I pass that value into blowfish build, it chokes because it is just using gcc, not gcc++. 

The "standard" flags we specify like "-fPIC -O2 -Wall" are already hard-coded in the makefile.  (omit-fram-pointer option is also hard-coded, though it specifies opposite of our standard option.)

In this case, it seems only this macosx flag is needed, but it makes sense to centralize it, in case the version ever needs to be modified.

patches/crypt_blowfish-makefile-cflags.patch
```diff
@@ -5,7 +5,7 @@
  LD = $(CC)
  RM = rm -f
 -CFLAGS = -fPIC -W -Wall -Wbad-function-cast -Wcast-align -Wcast-qual -Wmissing-prototypes -Wstrict-prototypes -Wshadow -Wundef -Wpointer-arith -O2 -fomit-frame-pointer -funroll-loops
-+CFLAGS = $(CFLAG_EXTRA) -fPIC -W -Wall -Wbad-function-cast -Wcast-align -Wcast-qual -Wmissing-prototypes -Wstrict-prototypes -Wshadow -Wundef -Wpointer-arith -O2 -fomit-frame-pointer -funroll-loops
++CFLAGS = $(EXTRA_CFLAGS) -fPIC -W -Wall -Wbad-function-cast -Wcast-align -Wcast-qual -Wmissing-prototypes -Wstrict-prototypes -Wshadow -Wundef -Wpointer-arith -O2 -fomit-frame-pointer -funroll-loops
```
mbautin, line 8: Still I am not sure why we need this dependency to behave differently from other dependencies. Can we just rename EXTRA_CFLAGS to CFLAGS, so the behavior of this would be that it would append to the existing value of CFLAGS?

yb_build_thirdparty_main.py
```diff
@@ -631,6 +632,7 @@ def init_flags(self):
             self.libs += ["-lc++", "-lc++abi"]
             # Build for macOS Mojave or later. See https://bit.ly/37myHbk
             self.cxx_flags.append("-mmacosx-version-min=10.14")
+            self.extrac_flags.append("-mmacosx-version-min=10.14")
```
mbautin, line 635: We are still duplicating this for no reason. Also we are missing other flags that other dependencies get through `self.compiler_flags` in crypt_blowfish. We should just make the crypt_blowfish build file accept an externally set value of CFLAGS, and append its own flags to it.

svarnau: Ah, if we put tis in compiler_flags, it will go into both make macros. I missed that common factor in the code.

PR #34, opened 2020-07-02
Add libedit to dependencies

svarnau:
yugabute/yugabyte-db#4946


build_definitions/libedit.py
```diff
@@ -0,0 +1,31 @@
...
+import sys
+
+sys.path.append(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
+
+from build_definitions import *
+
+class LibEditDependency(Dependency):
+    def __init__(self):
+        super(LibEditDependency, self).__init__(
+                'libedit', '20191231-3.1', 'https://thrysoee.dk/editline/libedit-{0}.tar.gz',
```
mbautin, line 25: It would be better to mirror libedit to our github.
https://docs.github.com/en/github/creating-cloning-and-archiving-repositories/duplicating-a-repository

svarnau: This seems like a good idea. I was unable to find an original source repo for libedit. The source is released as tarballs.  I found several repos on github that had copied the code for their own purpose, often older / modified versions. I also found some that put some automation around pulling from the original URL (as this change-set does). Even found one pulling the release tar files into a repo. 

I can try mirroring one of these other downstream repos, but did not find any real well maintained. Or I can set up our own manual mirror repo for libedit. Or I can set up subdirectory in this repo for checked in code to act as an archive of releases we care about.

Thoughts?

PR #37, opened 2020-08-01
filter dynamic libraries

slavestys:
Fix error:
$HOME/yugabyte-db/thirdparty/installed/uninstrumented/lib/pkgconfig/libtcmalloc_debug.pc:
Bad path: 	не является динамическим исполняемым файлом

### DIFF
--- a/yb_build_thirdparty_main.py
+++ b/yb_build_thirdparty_main.py
@@ -1048,7 +1048,7 @@ class LibTestLinux(LibTestBase):
     def good_libs(self, file_path):
         try:
             libout = subprocess.check_output(['ldd', file_path],
-                                             stderr=subprocess.STDOUT).decode('utf-8')
+                                             stderr=subprocess.STDOUT, env={'LC_ALL': 'en_US.UTF-8'}).decode('utf-8')
         except subprocess.CalledProcessError as ex:
             if ex.returncode > 1:
                 log("Unexpected exit code %d from ldd, file %s", ex.returncode, file_path)


slavestys:
Message is "не является динамическим исполняемым файлом". Not "not a dynamic executable"

svarnau:
The original post in slack sait the message was:
Bad path: 	**it is** not a dynamic executable file

I took it to mean that the existing regular expression did not match because of the extra "it is" words not in the pattern. But maybe that is an artifact of translation?

slavestys:
I just translated it into English. To make it clearer.
I fixed the commit. Please take a look.

svarnau:
Thanks. This is much more precise way to determine if we should be examining a particular file. Was this code (partially) imported from somewhere else? There are no comments indicating.

slavestys:
No. The file Format is quite simple.

slavestys:
But I didn't fully support it.

mbautin:
@slavestys Thank you for your contribution!

I have a few concerns regarding the proposed solution:
- Hard-coding low-level details about ELF file format.
- I think this kind of filtering is applied on both Linux and macOS, but it is not appropriate on macOS and will silently skip validation there.

How about just setting locale to ensure the error messages always show up in English?

```
23:45 ~ $ ( export LC_ALL=ru_RU.UTF-8; ldd /etc/hosts )
ldd: предупреждение: у вас нет прав на выполнение `/etc/hosts'
	не является динамическим исполняемым файлом
23:46 ~ $ ( export LC_ALL=en_US.UTF-8; ldd /etc/hosts )
ldd: warning: you do not have execution permission for `/etc/hosts'
	not a dynamic executable
```

slavestys:
Yes, this is how it should work. Should I correct the pull request?

svarnau:
> Yes, it should work. Should I correct the pull request?

Yes, that'd be great.

slavestys:
I fixed the commit.

PR #40, opened 2020-10-13
Refactor and add support for Clang 10 and GCC 8 (devtoolset-8)

mbautin:
Working towards enabling the following compilers:
- GCC 8 in devtoolset-8 on CentOS 7
- Custom-built Clang 10.0.1 on CentOS 7 (build scripts here: https://github.com/yugabyte/build-clang)

Also refactoring, cleanup, adding type safety annotations and pycodestyle checking.


svarnau (approved):
Lot of changes here, but I don't see any problems.
Just the automated test issues need to be addressed.

PR #44, opened 2020-11-06
Support Clang 10, 11 on CentOS 7 and Ubuntu 18.04, GCC 8, 9 devtoolset compilers on CentOS, major refactoring

mbautin:
(3rd pull request, submitting for CircleCI / GitHub Actions testing.)


svarnau (commented):
Wow, so much work to re-vamp this. The code looks good. 

I think the README and LICENSE files need to be updated.

So each of the new compiler types is a separate build & release? So instead of one thirdparty download per platform, there may be a different download with the various devtoolset versions?  This means that the yugabyte-db repo will need many more thirdparty URL files?

PR #47, opened 2021-01-16
Fix macOS builds; file system layout refactoring; license report

mbautin:
- Do not allow using CMake 3.19.1 because it fails to build libcds due to the issue https://gist.githubusercontent.com/mbautin/0fc792966b14ef8545379a392d5556ff/raw ( https://gitlab.kitware.com/cmake/cmake/-/issues/21529 ).
- Introduce a new FileSystemLayout class that encapsulates the knowledge about the layout of the third-party repository source/build/installation tree.
- Add an ability to generate a license report for the third-party dependencies. This is needed for the open source disclosure file generation. This report also includes copyright notices that are found within the source code directories in files such as LICENSE, COPYING, NOTICE, etc.
- Update Docker images for CentOS 7 and Ubuntu 18.04 to those built on 12/14/2020. (See https://github.com/yugabyte/build-infra for how to find latest images.)
- Do not check dynamic libraries in the --download-extract-only mode.


build_and_release.sh
```diff
@@ -5,6 +5,43 @@ set -euo pipefail
 # shellcheck source=./yb-thirdparty-common.sh
 . "${BASH_SOURCE%/*}/yb-thirdparty-common.sh"
 
+install_cmake_on_macos() {
```
svarnau, line 8: This seems to be done  regardless of what version currently exists. Does this work on mac dev system if run more than once?

Does it work in yugabyte-db  tree if dev decides not to download thirdparty, but build it? And what if they don't have sudo access?

mbautin: This script is only invoked in the CI/CD path, not as part of YugabyteDB build.

mbautin: Actually I've removed this, as CMake 3.19.2 and onwards works.

PR #48, opened 2021-01-20
Automatically download compiler toolchain

mbautin:
- Add an option to automatically download compiler toolchain (Linuxbrew or LLVM) in Python scripts.
- Create files containing compiler toolchain URL and installation path in the resulting thirdparty tarball. This generalizes the concept we've been using for Linuxbrew to the LLVM toolchain as well.
- Do not attempt to upload releases to GitHub when building a pull request, even if GITHUB_TOKEN is specified.
- Actually start using the existing compiler version identification logic. Explicitly disallow GCC versions older than 5.5.0. Failures with GCC 4.8 that is default on CentOS 7 are non-obvious.
- Install CMake 3.19.2 on macOS before doing the build in case the installed version turns out to be 3.19.1, which is broken. The issue with CMake 3.19.1 is show in https://gist.githubusercontent.com/mbautin/0fc792966b14ef8545379a392d5556ff/raw and https://gitlab.kitware.com/cmake/cmake/-/issues/21529.


svarnau (approved):
Except for remaining test failure, this looks pretty good.

PR #56, opened 2021-04-01
FOSSA integration and other improvements

mbautin:
FOSSA integration: generate a fossa_modules.yml file that can be used to generate raw modules in .fossa.yml corresponding to C/C++ dependencies. We also include URLs and SHA256 checksums of the exact dependencies used, allowing scripts in the yugabyte-db repo to download dependency tarballs and generate a final .fossa.yml file.

Other changes:
- Move tarball creation and upload to Python.
- Use the codecheck module ( https://pypi.org/project/codecheck/ ) for checking for errors in Python and Bash code.


svarnau:
The python checks are failing, but code all looks pretty reasonable to me.

svarnau (approved):

PR #57, opened 2021-04-30
[yugabyte/yugabyte-db#7778] Update OpenSSL to 1.1.1k

spolitov:


thirdparty_src_checksums.txt
```diff
@@ -35,8 +35,7 @@ c7ccb735c37b4ec470f66a6c35fbae4f029c0f88038f6977180b1a8ddc255637  llvm-10.0.1.ta
 1bcc9b285074ded87b88faaedddb88e6b5d6c331dfcfb57d7f3393dd622b3764  llvm7-7.1.0.tar.xz
 c48450d27524c2e5856997133e059e3cf9909241110a6e21ad278890ac425afc  lz4-r130.tar.gz
 86af21a2db98f8328a1cf98bf725b17d1c3d9c6cf0a8a37144ad55406c3c252b  openldap-2_4_54.tar.gz
-ecd0c6ffb493dd06707d38b14bb4d8c2288bb7033735606569d8f90f89669d16  openssl-1.0.2u.tar.gz
-ddb04774f1e32f0c49751e21b67216ac87852ceb056b75209af2443400636d46  openssl-1.1.1g.tar.gz
```
bmatican, line 39: did we have both, because we were transitioning? ie: is it ok to just have 1.1.1 now?

python/build_definitions/openssl.py
```diff
@@ -31,7 +31,7 @@ def build(self, builder: BuilderInterface) -> None:
         if is_mac():
             # On macOS x86_64, OpenSSL 1.0.2 fails to detect the proper architecture.
             configure_cmd = [
-                '/bin/bash', './Configure', 'darwin64-x86_64-cc'] + common_configure_options
```
bmatican, line 34: was this not necessary in the first place, or did something change?

thirdparty_src_checksums.txt
spolitov: I think usage of 1.0.2u was incomplete

python/build_definitions/openssl.py
spolitov: Because 1.1.1k Configure is written in Perl
```#! /usr/bin/env perl```

PR #58, opened 2021-05-02
Support Clang 7 without Linuxbrew and update Clang 12 to a released version

mbautin:
- Support building with Clang 7 without Linuxbrew. Adding a new "llvm7" toolchain.
- Update the LLVM12 toolchain to use the official LLVM/Clang 12 release (not a release candidate).


python/yugabyte_db_thirdparty/builder.py
```diff
@@ -592,6 +604,12 @@ def init_linux_clang7_flags(self, dep: Dependency) -> None:
             self.compiler_flags.append('--gcc-toolchain={}'.format(
                 self.compiler_choice.get_linuxbrew_dir()))
 
+        if self.toolchain and self.toolchain.toolchain_type == 'llvm7':
+            # This is needed when building with Clang 7 but without Linuxbrew.
+            # TODO: this might only be needed due to using an old version of libunwind that is
+            # different from libunwind included in the LLVM 7 repository. Just a hypothesis.
+            self.ld_flags.append('-lgcc_s')
+
```
spolitov, line 612: Could we create a separate function for this workaround and use it here and in i4cu bulid?

.circleci/config.yml
```diff
@@ -71,20 +71,21 @@ workflows:
...
       - build:
-          name: centos7-clang12rc1
-          archive_name_suffix: centos7-clang12rc1
+          name: centos7-clang12
+          archive_name_suffix: centos7-clang12
           docker_image: yugabyteci/yb_build_infra_centos7:v2020-12-14T01_12_21
-          build_thirdparty_args: >-
-            --single-compiler-type=clang
-            --toolchain=llvm12rc1
+          build_thirdparty_args: --toolchain=llvm12
```
spolitov, line 88: Do we need clang12 related changes in this PR?

thirdparty_src_checksums.txt
```diff
@@ -32,15 +32,18 @@ af02b440ce3afe08db490eee1a685ca1e9d276d56a9ed2f2ea1ade2245426e67  libunwind-1.1a
 c7ccb735c37b4ec470f66a6c35fbae4f029c0f88038f6977180b1a8ddc255637  llvm-10.0.1.tar.gz
 8ad4ddbafac4f2c8f2ea523c2c4196f940e8e16f9e635210537582a48622a5d5  llvm-11.0.0.tar.gz
 9c7ad8e8ec77c5bde8eb4afa105a318fd1ded7dff3747d14f012758719d7171b  llvm-11.0.1.tar.gz
```
spolitov, line 34: Should we remove unused checksums from this file?

spolitov (changes requested):

python/yugabyte_db_thirdparty/builder.py
mbautin: Actually, this is probably not needed here. The flag in builder.py will apply to all dependencies.

thirdparty_src_checksums.txt
mbautin: I think it is fine to leave them here for now. We can bulk-remove them later when it becomes clear that we will never go back to certain old versions.

.circleci/config.yml
mbautin: I think it is perfectly fine to also do Clang 12 changes here. Clang 12 has been released and I am replacing the release candidate with the official release build.

svarnau (approved):
LGTM

PR #76, opened 2021-07-06
Changed cqlsh version to 3.10-yb-8

OlegLoginov:
Updated: Reduce the default INGESTRATE to 1k
https://github.com/yugabyte/cqlsh/pull/8


thirdparty_src_checksums.txt
```diff
@@ -4,9 +4,7 @@
 e97dc472aae52197a4d5e0185eb8f9e04d7575d2dc2b12194ddc768e0f8a846d  cfe-7.1.0.tar.xz
 1ce0042c48ecea839ce67b87e9739cf18e7a5c2b3b9a36d177d00979609b6451  clang-tools-extra-7.1.0.tar.xz
 057bdac0581215b5ceb39edfd5bbef9eb79578f16a8908349f3066251fba88d8  compiler-rt-7.1.0.tar.xz
-035314f04e334bb3d7b3f57330af39d1e3d2070b8d2f8ffe5d873b37e98443a5  cqlsh-3.10-yb-5.tar.gz
-d55ac6565b8694f3861b6dbb323a195596b9b65b68460d57279959bd8c9a950f  cqlsh-3.10-yb-6.tar.gz
-c834e264530f805fe6068dd5c3339b9da3c55d9ed02cc58dd0436e83d8c1e173  cqlsh-3.10-yb-7.tar.gz
```
m-iancu, line 9: @OlegLoginov Do we know why we previously kept the old versions around here? 
Is there a reason for that?
cc @mbautin

OlegLoginov: > @OlegLoginov Do we know why we previously kept the old versions around here?
> Is there a reason for that?

No reason.
Previous version referencing to v7 only (v5 & v6 are not referenced):
```
            version='3.10-yb-7',
```

m-iancu (approved):
LGTM!

PR #79, opened 2021-07-30
Fix build using official Clang on CentOS 8 under aarch64

mbautin:
Adding a missing library path when building libc++ so that it can find libc++abi in the lib directory that is common both to libc++ and libc++abi. Without this, the build does not work on CentOS 8 under aarch64 using the official clang package (clang 11.0.0).

Also refactoring/cleanup:
- Using the compiler-identification module for identifying compiler version.
- Do not build LLVM 7 as part of this repository (while still allowing to build LLVM7's libc++ for ASAN/TSAN).
- Update requirements_frozen.txt.


svarnau (approved):

jharveysmith (approved):

jharveysmith:
Should we add the arch to the name of the build artifacts and start building centos8-arch64?

PR #81, opened 2021-08-24
Add x86_64 AlmaLinux 8 builds

mbautin:
- Do not attempt to build code with ASAN on GCC. We have not been using ASAN on GCC because it does not support suppressions, or did not support them in a reasonably recent version, such as 8 or 9. Not sure if this is accurate but, in any case, we should focus on upgrading the Clang version we use with ASAN/TSAN first. A relevant bug: https://bugzilla.redhat.com/show_bug.cgi?id=1684182
- Use a more universal way for finding the Clang runtime library directory that works with Clang 11 shipped as part of AlmaLinux 8.
- Fix devtoolset support on AlmaLinux. The previous approach based on the env command does not work because env shows definitions of Bash functions and we cannot parse them. Instead, just print the variables that we need.
- Add a `--expected-major-compiler-version` argument to validate that the major compiler version we are declaring in the archive name matches the version we are using.
- Use the new sys-detection module ( https://pypi.org/project/sys-detection/ ) to come find the correct toolchain archive, and wherever we need to distinguish between macOS and Linux and RedHat vs. non-RedHat, etc.
- Support our custom-built LLVM 11 and LLVM 12 toolchains on AlmaLinux 8.


svarnau:
Why don't we just use same binaries for centos8 and almalinux8 ?

mbautin:
@svarnau 
> I think the other improvements are good, but seems somewhat confusing to build and publish two different packages that are supposed to be binary compatible. If we add, almalinux8, then why not stop building centos8?

I think it remains to be verified that the binary compatibility works as advertised. For now I would prefer to build separate packages. In the future we might be able to reuse one of them (perhaps AlmaLinux 8) across all RHEL 8 compatible distros. Also with CentOS 8 there is CentOS 8 vs. CentOS 8 Stream confusion.

svarnau:
> @svarnau
> 
> > I think the other improvements are good, but seems somewhat confusing to build and publish two different packages that are supposed to be binary compatible. If we add, almalinux8, then why not stop building centos8?
> 
> I think it remains to be verified that the binary compatibility works as advertised. For now I would prefer to build separate packages. In the future we might be able to reuse one of them (perhaps AlmaLinux 8) across all RHEL 8 compatible distros. Also with CentOS 8 there is CentOS 8 vs. CentOS 8 Stream confusion.

We have already been using the centos8-built dependencies with success on alma8, so I have some confidence the binary compatibility is good between these EL8 family distributions. There does seem some confusion with Stream. It would be fine to swiitch to build on AlmaLinux. I think in that case, we can drop the centos8 flavor.

svarnau (approved):

mbautin:
Sounds good, thanks for the review @svarnau! We can drop the CentOS 8 build in a later commit.

PR #85, opened 2021-09-28
fossa: generate fossa-deps.json config

svarnau:
This replaces the fossa-modules.yml file that was pulled into fossa
analysis in the yugabyte-db repo. This assumes analyzing this
yugabyte-db-thirdparty repo directly.

Test:
  ./build_thirdparty.sh  --download-extract-only
  fossa analyze -o

To-Do: Add to release workflow

### DIFF
--- a/.fossa.yml
+++ b/.fossa.yml
@@ -1,0 +1,13 @@
+
+version: 3
+
+project:
+  id: yugabyte-db-thirdparty
+
+revision:
+  branch: master
+
+# Look only at fossa-deps.yml
+paths:
+  only:
+    - None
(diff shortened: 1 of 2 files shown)


.fossa.yml
```diff
@@ -0,0 +1,13 @@
+
```
mbautin, line 1: Why do we need this file? Where is it being referenced from?

mbautin (approved):

.fossa.yml
svarnau: Running fossa analyze references this as the main config file. If there is none, it will try to analyze everything in the repo. We only want to analyze the fossa-deps.yml file, so we need a config file to tell it to ignore all the other files.

PR #90, opened 2021-11-21
tcmalloc fix for macOS Monterey; do not use Homebrew libuv; fix Boost flags; fix min macOS version

mbautin:
https://github.com/yugabyte/yugabyte-db/issues/10533 (incorrectly using libuv from Homebrew):
- Explicitly specify libuv search path when building Cassandra C++ driver. This prevents us from using Homebrew-provided libuv even if it is installed.
- Do not allow Homebrew library paths to appear in `otool -L` output when checking dynamic libraries. This enforces that the resulting thirdparty package does not depend on any Homebrew libraries, e.g. libuv.

https://github.com/yugabyte/yugabyte-db/issues/10448 (tcmalloc issue on macOS Monterey):
Update gperftools with a patch that fixes tcmalloc on macOS Monterey.
Upstream gperftools issue: gperftools/gperftools#1312, fix: https://github.com/gperftools/gperftools/pull/1315

https://github.com/yugabyte/yugabyte-db/issues/10661 (incorrectly setting min macOS version):
Add validation that libraries are built for the correct target version of macOS by examining the output of `otool -l`.
As part of looking into this, I found out that Boost previously did not even use the compiler and linker flags that we were providing. Fixed this by specifying the `toolset=...` parameter to b2. Also fixed Boost build on x86_64 macOS (after the arm64 commit landed, we would always specify the arm64 architecture). Another dependency that required changes was crypt_blowfish (needed to pass `-mmacosx-version-min=...` in LDFLAGS and ASFLAGS). I noticed that specifying min macOS version as 10.14 (Mojave) does not work when building for arm64, so I specified min macOS version as 11.2 there for now. Also specifying minimum macOS version does not seem to work correctly when building on macOS 10.15 (Catalina) on GitHub Actions, but works well on Big Sur (11.x) so we will look into upgrading our build workers, and we may have to release a few packages manually.

Other improvements: 
- Specify patch_strip as 1 by default so we don't have to specify it everywhere.
- Add a --force flag to rebuild dependencies even if the system thinks it does not need to based on "stamp" files.
- Add a --delete-build-dir to delete a dependency's build directory before attempting to build that dependency.
- Always rebuild dependencies if the build directory is missing.
- Generate convenient shell scripts allowing to re-run CMake builds manually.
- Expand the "patch marker" file naming format to include the number of patches.
- Remove unused test data for compiler version detection (we are now using the https://pypi.org/project/compiler-identification/ module).
- The redis_cli dependency used to have custom logic to patch how libhiredis refers to itself on macOS. This patching belongs in hiredis.py, and we have a standard implementation for it (fix_shared_library_references) so the custom code from redis_cli.py is being deleted.


python/yugabyte_db_thirdparty/build_definition_helpers.py
```diff
@@ -24,6 +24,7 @@
     remove_path,
     copy_file_and_log
 )
+from yugabyte_db_thirdparty.arch import is_macos_arm64_build
```
spolitov, line 27: Do we need this one?

python/yugabyte_db_thirdparty/builder.py
```diff
@@ -77,6 +79,10 @@
     '-DTHREAD_SANITIZER',
 ]
 
+# We create a file named like this in each dependency's build directory, with all the relevant
+# environment variables that we set.
+DEPENDENCY_ENV_FILE_NAME = 'yb_dependency_env.sh'
```
spolitov, line 84: Do we need this extra file?
Would it be easier to write env vars directly to yb_build_with_cmake.sh?

spolitov (approved):

python/yugabyte_db_thirdparty/build_definition_helpers.py
mbautin: Yes, we include this file from each build definition file and we use this function in a couple of them.

python/yugabyte_db_thirdparty/builder.py
mbautin: We currently write the env var file for both configure- and CMake-based builds, and I think the environment-only file is still useful.

python/yugabyte_db_thirdparty/compiler_wrapper.py
```diff
@@ -31,8 +31,18 @@ def run(self) -> None:
...
+
         compiler_path_and_args = [real_compiler_path] + compiler_args
 
+        cmd_line_str = shlex_join(compiler_path_and_args)
+        # if (cmd_line_str.strip() != '/usr/bin/clang++ --version' and
+        #         not '-print-prog-name=' in cmd_line_str and
+        #         not '-dump' in cmd_line_str):
+        #     raise ValueError("Cmd line: %s" % cmd_line_str  )
+
         if use_ccache:
```
spolitov, line 46: Do we need this commentedout code?

python/yugabyte_db_thirdparty/compiler_choice.py
```diff
@@ -41,13 +41,15 @@ class CompilerChoice:
...
+    c_compiler_or_wrapper: Optional[str]
+    cxx_compiler_or_wrapper: Optional[str]
     single_compiler_type: Optional[str]
     compiler_prefix: Optional[str]
     compiler_suffix: str
     devtoolset: Optional[int]
     linuxbrew_dir: Optional[str]
     use_compiler_wrapper: bool
-    use_ccache: bool
+    use_compiler_argsache: bool
```
spolitov, line 52: argsache?

spolitov (approved):

python/yugabyte_db_thirdparty/compiler_wrapper.py
mbautin: Removed.

python/yugabyte_db_thirdparty/compiler_choice.py
mbautin: Reverted the typo.

PR #95, opened 2021-12-21
Support for Clang 12 build for Linuxbrew (to replace GCC 5.5)

mbautin:
Relevant issue: https://github.com/yugabyte/yugabyte-db/issues/10920

Using Clang 12 that is natively compiled for the host OS (currently, RHEL 7 or 8 compatible distros) to build third-party dependencies using the same x86_64 Linuxbrew archive that we are currently using for the production GCC 5.5 based build. This will allow us to replace GCC 5.5 with Clang 12 and use modern C++ features.

We achieve this by specifying `-nostdinc` and `-nostdinc++` and specifying all the include directories manually. We also specify Linuxbrew's library directories both as `-L` and `-Wl,-rpath`. 

Boost's build system would add `-I/usr/include` to the compiler command line for some reason. We are working around this by activating our compiler wrapper and removing that argument from the command line. We are also using the compiler wrapper to validate the set of actual include files used and to verify that no include files from `/usr/include` are used.

### DIFF
--- a/python/build_definitions/snappy.py
+++ b/python/build_definitions/snappy.py
@@ -29,11 +29,32 @@ class SnappyDependency(Dependency):
         self.patches = ['snappy-define-guard-macro.patch']
         self.post_patch = ['autoreconf', '-fvi']
 
+    def _disable_lzo2_library_in_test(self) -> None:
+        '''
+        Makes the snappy unit test not use the liblzo2 library. Sometimes the configure script will
+        pick up the library from a system directory when it should not.
+        '''
+        log("Removing HAVE_LIBLZO2 from config.h")
+        lines: List[str] = []
+        removed = False
+        with open('config.h') as input_file:
+            for line in input_file:
+                if line.strip() == '#define HAVE_LIBLZO2 1':
+                    removed = True
+                else:
+                    lines.append(line)
+        if not removed:
+            log("Warning: did not remove HAVE_LIBLZO2 from config.h")
+        with open('config.h', 'w') as output_file:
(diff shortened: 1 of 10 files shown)


python/build_definitions/snappy.py
```diff
@@ -29,11 +29,31 @@ def __init__(self) -> None:
...
+        with open('config.h', 'w') as output_file:
+            output_file.write('\n'.join(lines) + '\n')
+
     def build(self, builder: BuilderInterface) -> None:
         log_prefix = builder.log_prefix(self)
         builder.build_with_configure(
             log_prefix=log_prefix,
-            extra_args=['--with-pic']
+            extra_args=['--with-pic'],
+            post_configure_action=self._disable_lzo2_library_in_test
```
spolitov, line 56: Could we put comma to the end of this line?

python/yugabyte_db_thirdparty/builder.py
```diff
@@ -137,6 +139,30 @@ def __init__(self) -> None:
...
+
+        # We expect at most one Linuxbrew toolchain to be specified (handled by set_linuxbrew_dir).
+        for toolchain in toolchains:
+            if toolchain.toolchain_type == 'linuxbrew':
+                set_linuxbrew_dir(toolchain.toolchain_root)
+
+        if len(toolchains) == 1:
+            self.toolchain = toolchains[0]
+            return
+        if len(toolchains) != 2:
```
spolitov, line 154: Looks like we always expect llvm_linuxbrew in this case.

spolitov (approved):

PR #98, opened 2022-02-05
Clang LTO support

mbautin:
Adding an `--lto` flag with potential values `full` and `thin`, which results in passing `-flto=full` or `-flto=thin` to Clang compiler command line. The resulting static libraries can then be LTO-linked with YugabyteDB code.

Removing older GCC build types, CentOS 8 build types (replaced with AlmaLinux 8), and Clang 10 and Clang 11 build types. Adding thin and full LTO build types for AlmaLinux 8, with and without Linuxbrew.

Also fixing a bug for remote build support: `--remote-...` command-line flags should not be passed to the remote build.

The crypt_blowfish library now produces a shared library in addition to the static library (changes: https://github.com/yugabyte/crypt_blowfish/compare/v1.0.0..v1.3.2).


python/yugabyte_db_thirdparty/clang_util.py
```diff
@@ -58,3 +60,31 @@ def get_clang_include_dir(clang_executable_path: str) -> str:
             return include_dir
     raise ValueError(
         f"Could not find a directory from {library_dirs} that has an 'include' subdirectory.")
+
+
+def create_llvm_tool_dir(clang_path: str, tool_dir_path: str) -> bool:
+    """
+    Create a directory with tools that are named like
```
svarnau, line 67: named like what?

svarnau (approved):
incomplete comment?

python/yugabyte_db_thirdparty/clang_util.py
mbautin: Fixed.

PR #100, opened 2022-02-15
Adding build definition for krb5

nathanhjli:
Adding the build definition for krb5 to support GSSAPI/Kerberos authentication


python/build_definitions/krb5.py
```diff
@@ -24,7 +24,7 @@ def __init__(self) -> None:
             'krb5',
             '1.19.2',
             'https://kerberos.org/dist/krb5/1.19/krb5-{0}.tar.gz',
-            BUILD_GROUP_INSTRUMENTED)
+            BUILD_GROUP_COMMON)
```
mbautin, line 27: I think this should still be in the instrumented group.

nathanhjli:
@mbautin To resolve these build failures, do you think we'd need to create our own fork of krb5? I'm trying to look at the command that failed and that seems to originate from `shlib.conf`. I think that command needs to include `-fsanitize=address` somehow but that doesn't seem to be configurable.

PR #101, opened 2022-02-22
Update boost to 1.78

spolitov:
Removed outdated patches: 
  boost-1-69-remove-pending-integer_log2-include.patch - already included in new boost versions
  boost-1-69-mac-compiler-flags.patch - not sure why it was needed, since we don't build boost with GCC

Updated to new boost version:
  boost-1-78-add-arm64-instruction-set.patch - patched files were updated, so need to adjust patch also.


python/build_definitions/boost.py
```diff
@@ -48,10 +48,8 @@ def __init__(self) -> None:
             license='Boost Software License 1.0')
         self.dir = '{}_{}'.format(self.name, self.underscored_version)
         self.copy_sources = True
-        self.patches = ['boost-1-69-remove-pending-integer_log2-include.patch',
-                        'boost-1-69-mac-compiler-flags.patch']
-        if is_macos_arm64_build():
-            self.patches.append('boost-1-69-add-arm64-instruction-set.patch')
+        # if is_macos_arm64_build():
```
mbautin, line 51: Remove commented-out code.

mbautin:
It would be good to add a description of what's happening with custom architecture flags / custom patches, and what is becoming unnecessary with the updated version of Boost.

mbautin:
LGTM. Reminder: please include the message from the PR into the commit message.

PR #106, opened 2022-03-30
Update zlib to 1.2.12

def-:
"Fix a deflate bug when using the Z_FIXED strategy that can result in out-of-bound accesses." sounds serious.


svarnau:
Builds failing -- looks like some compile options not compatible with clang

def-:
For some reason the linking command seems to use `cc` instead of `/opt/yb-build/llvm/yb-llvm-v11.1.0-yb-1-1633099975-130bd22e-centos7-x86_64/bin/clang`. I'm trying to find the reason.

def-:
Seems like my workaround also didn't work. Not sure why the linking is still not working.

svarnau:
[zlib (common)] cc: error: unrecognized command line option '-rtlib=compiler-rt'
[386](https://github.com/yugabyte/yugabyte-db-thirdparty/runs/5757644355?check_suite_focus=true#step:3:386)
[zlib (common)] cc: error: unrecognized command line option '-fuse-ld=lld'

mbautin:
@def- are you sure the zlib build system is using the correct compiler? Clang definitely supports these options -- maybe zlib is now using gcc?

def-:
@mbautin The Makefile for zlib seems to be doing something strange. As I wrote previously:

> For some reason the linking command seems to use `cc` instead of `/opt/yb-build/llvm/yb-llvm-v11.1.0-yb-1-1633099975-130bd22e-centos7-x86_64/bin/clang`. I'm trying to find the reason.

I'm not sure if we are supposed to patch the Makefile now or if there is a better workaround:
```
CC=cc
[...]
LDSHARED=$(CC)
```
This appears to end up with CC being set correctly, but LDSHARED being `cc`

mbautin:
@def- Yes, we need to ensure that we can control the compiler driver used for both compiling and linking. Apparently the previous version of zlib we were using did not have this problem. We might need to create our own fork of zlib under the yugabyte organization on GitHub. Also see https://github.com/yugabyte/yugabyte-db-thirdparty#modifying-build-definition-for-a-dependency.

def-:
Failure is unrelated, zlib build succeeded: `Finished building zlib (common)` Instead llvmorg hash fails, can't reproduce locally:
```
yugabyte_db_thirdparty.custom_logging.FatalError: File '/opt/yb-build/thirdparty/yugabyte-db-thirdparty-v20220331062353-5daa8ddaa7-ubuntu2004-x86_64-clang11/download/llvm-11.1.0.tar.gz' has wrong checksum after downloading from 'https://github.com/llvm/llvm-project/archive/llvmorg-11.1.0.tar.gz'. Has 8e8052a0c2d1b3568998cc84bea4b37c23a7db71d547cf8f34b6b109f2f55a30, but expected: 53a0719f3f4b0388013cfffd7b10c7d5682eece1929a9553c722348d1f866e79
```

I compared zlib 1.2.11 and 1.2.12 and the Makefiles had no relevant changes. My guess is that we changed the build after having already built zlib 1.2.11

def-:
Ok, can reproduce it locally. zlib 1.2.12 changed the way its cc detection in `./configure` works.

def-:
Thanks for the help @mbautin , found the culprit in `./configure`, made a patch for us and will also try to get this fixed upstream. cc is not being set to CC if CC is set.

svarnau (approved):

dscho:
> Failure is unrelated, zlib build succeeded: `Finished building zlib (common)` Instead llvmorg hash fails, can't reproduce locally:
> 
> ```
> yugabyte_db_thirdparty.custom_logging.FatalError: File '/opt/yb-build/thirdparty/yugabyte-db-thirdparty-v20220331062353-5daa8ddaa7-ubuntu2004-x86_64-clang11/download/llvm-11.1.0.tar.gz' has wrong checksum after downloading from 'https://github.com/llvm/llvm-project/archive/llvmorg-11.1.0.tar.gz'. Has 8e8052a0c2d1b3568998cc84bea4b37c23a7db71d547cf8f34b6b109f2f55a30, but expected: 53a0719f3f4b0388013cfffd7b10c7d5682eece1929a9553c722348d1f866e79
> ```
> 
> I compared zlib 1.2.11 and 1.2.12 and the Makefiles had no relevant changes. My guess is that we changed the build after having already built zlib 1.2.11 and didn't trigger any rebuild of zlib since we added those linking flags.

For the record, we ran into the same issue, and [figured out](https://github.com/microsoft/homebrew-git/issues/102#issuecomment-4651847977) that this is the SHA-256 of GitHub's Unicorn page!

PR #107, opened 2022-04-03
Update LLVM URLs

mbautin:


def-:
@spolitov amzn2-aarch64 for clang11 is available now in case you still need it.

def- (approved):
I assume we only use the yb-1 if we had to change something and yb-0 is implicit.

mbautin:
Right, the only commit we needed to cherry-pick into LLVM 11/12/13 branches is already present in LLVM 14.0.0.

python/build_definitions/protobuf.py
```diff
@@ -21,8 +21,9 @@
 class ProtobufDependency(Dependency):
     def __init__(self) -> None:
         super(ProtobufDependency, self).__init__(
-            'protobuf', '3.5.1',
-            'https://github.com/google/protobuf/releases/download/v{0}/protobuf-cpp-{0}.tar.gz',
+            'protobuf',
+            '3.5.1-yb-1',
+            'https://github.com/yugabyte/protobuf/archive/refs/tags/v{0}.tar.gz',
```
svarnau, line 26: It would be nice to see a bit more rationale for YB-specific version os protobuf, either here or in the yugabyte-fork of that repo.

svarnau (approved):

svarnau:
Note that the clang14 LTO builds are failing in ASAN

python/build_definitions/protobuf.py
mbautin: Explained in the description.

PR #108, opened 2022-04-09
[yugabyte/yugabyte-db#12089] Add ncurses dependency; upgrade to OpenSSL 1.1.1n

mbautin:
Issue: https://github.com/yugabyte/yugabyte-db/issues/12089

This is one of the few remaining dependencies we have on system packages. It is because of ncurses and glibc that we can't take code built for CentOS 7 and run it e.g. on RHEL 8. If we build all of dependencies ourselves, we will be able to produce a "universal Linux" third-party library archive for each architecture.

Found that we were duplicating include flags in the CPPFLAGS (meaning "C preprocessor flags") environment variable as well as in CXXFLAGS and CFLAGS variables, which causes failures to build with LLVM with custom glibc (i.e. while using Linuxbrew). Refactored the framework so that each dependency class can specify whether to use the CPPFLAGS variable by overriding the `use_cppflags_env_var` method.

Also upgrading to OpenSSL 1.1.1n, and cleaning up the code by renaming "llvm1x" to simply "llvm" (we don't support LLVM versions less than 10 anymore).


python/yugabyte_db_thirdparty/arch.py
```diff
@@ -40,6 +40,9 @@ def get_target_arch() -> str:
         return g_target_arch
 
     g_target_arch = os.getenv('YB_TARGET_ARCH')
+    if g_target_arch is None:
+        g_target_arch = platform.machine()
+
     if g_target_arch not in MACOS_CPU_ARCHITECTURES:
         raise ValueError("Unsupported value of YB_TARGET_ARCH on maOS: %s" % g_target_arch)
```
svarnau, line 47: Not your change, but there is a line 47 typo: maOS

svarnau (approved):
LGTM

python/yugabyte_db_thirdparty/builder.py
```diff
@@ -406,7 +419,7 @@ def add_include_path(self, include_path: str) -> None:
             log("Adding an include path: %s", include_path)
         cmd_line_arg = f'-I{include_path}'
         self.preprocessor_flags.append(cmd_line_arg)
-        self.compiler_flags.append(cmd_line_arg)
+        # self.compiler_flags.append(cmd_line_arg)
```
nathanhjli, line 422: Should we be commenting these out or removing them entirely? For here and all the other occurrences.

nathanhjli (approved):

PR #111, opened 2022-05-05
GCC 11 support; RapidJSON bug fix; add Ubuntu 22.04; update OpenSSL to 1.1.1o

mbautin:
- Adding support for GCC 11 devtoolset on RHEL-like OSes.
- Adding Ubuntu 22.04 builds with Clang 12, 13, 14, and GCC 11.
- Pulling in a bug fix that causes RapidJSON to fail to build in some cases (yugabyte/yugabyte-db#12365)
- Upgrading OpenSSL to 1.1.1o
- Moving the logic to determine LLVM package download URL for a particular OS/version/architecture combination to a separate Python module (https://github.com/yugabyte/llvm-installer, https://pypi.org/project/llvm-installer/).


d-uspenskiy (approved):
LGTM

svarnau (approved):

PR #114, opened 2022-05-05
Add libkeyutils and libverto, enable static linking for krb5, remove libatomic dependency when possible

mbautin:
- libkeyutils and libverto are dependencies of krb5 and if we don't build them, the corresponding system dependencies will be pulled in.
- Previously, krb5 was only build in shared mode. Linking with krb5 would result in some other shared libraries being pulled in, and as a result yb-tserver's LTO-enabled build would link with some libraries twice, both as static and shared library, which is undesirable.
- During checking shared library dependencies, we are now only allowing a small number of system libraries (those in /usr/lib64), primarily those that are part of glibc, as well as libgcc_s.
- Also during checking shared library dependencies, we are now removing unused "needed libraries" as shown by `ldd -u` Currently we only remove libatomic in this way. Removing other dependencies like libgcc_s sometimes results in shared libraries on which the dynamic linker throws errors, so we have to be cautious.


svarnau (approved):
LGTM

python/yugabyte_db_thirdparty/cmd_line_args.py
```diff
@@ -25,6 +27,12 @@
 from build_definitions import BUILD_TYPES
 
 
+INCOMPATIBLE_ARGUMENTS: Dict[str, Set[str]] = {
+    'toolchain': set(['devtoolset', 'compiler_prefix', 'compiler_suffix']),
```
def-, line 31: nitpick, can create sets using {} literal without converting from list: `{'devtoolset', 'compiler_prefix', 'compiler_suffix'}`

python/yugabyte_db_thirdparty/dependency.py
```diff
@@ -33,6 +33,10 @@ class Dependency:
     mkdir_only: bool
     archive_name: Optional[str]
 
+    # For dependencies built with configure/autotools, where out-of-source build is not possible,
+    # this tells the initial step to create separate build directories for shared and static builds.
+    shared_and_static: bool
+
     # Enforce the use of lld linker on Clang 14 and later by appending it to the linker flags in
     # the compiler wrapper.
     enforce_lld_in_compiler_wrapper: bool
```
def-, line 42: unused

python/yugabyte_db_thirdparty/library_checking.py
```diff
@@ -22,15 +27,71 @@
...
+
 def compile_re_list(re_list: List[str]) -> Any:
     return re.compile("|".join(re_list))
 
 
+def get_needed_libs(file_path: str) -> List[str]:
+    return capture_all_output(
+        ['patchelf', '--print-needed', file_path],
+        allowed_exit_codes={1},
+        extra_msg_on_nonzero_exit_code=f"Warning: could not determine libraries directly "
```
def-, line 91: no need for `f` formatting

python/yugabyte_db_thirdparty/library_checking.py
```diff
@@ -110,18 +189,34 @@ def run(self) -> None:
...
             log("No problems found with library dependencies.")
 
+    def before_checking_all_files(self) -> None:
+        pass
+
+    def check_all_files(self) -> bool:
+        success = True
+        for file_path in self.files_to_check:
+            if not self.check_libs_for_file(file_path):
+                success = False
```
def-, line 217: Could return here already instead of iterating further

python/yugabyte_db_thirdparty/library_checking.py
```diff
@@ -110,18 +189,34 @@ def run(self) -> None:
...
         if not test_pass:
             fatal(f"Found problematic library dependencies, using tool: {self.tool}")
         else:
             log("No problems found with library dependencies.")
 
+    def before_checking_all_files(self) -> None:
+        pass
+
+    def check_all_files(self) -> bool:
+        success = True
```
def-, line 214: (matter of taste, but short circuits automatically) `return all(map(self.check_libs_for_file, self.files_to_check))`

def- (approved):
Only some minor suggestions

python/yugabyte_db_thirdparty/library_checking.py
mbautin: This is on purpose, to print all errors.

python/yugabyte_db_thirdparty/library_checking.py
mbautin: See my earlier comment. I don't want to return early from this loop.

PR #115, opened 2022-05-13
SEC-76 Snyk plugin integration in github workflow

bkms2018:
https://yugabyte.atlassian.net/browse/SEC-76

This PR adds snyk scanning to code without influencing the current test workflow. The code scan results are updated at each scan on snyk dashboard here - https://app.snyk.io/org/yugabyte-inc/project/cd6f8eb6-8590-4e56-912a-f59f12725f83

### DIFF
--- a/build_thirdparty.sh
+++ b/build_thirdparty.sh
@@ -62,7 +62,7 @@ if [[ $* == *--snyk* ]]; then
     exit $rc
   fi
 
-  ./snyk test "$homedir/src" --unmanaged
+  ./snyk monitor "$homedir/src" --unmanaged
   rc=$?
   rm -f ./snyk
 fi
--- a/build_thirdparty.sh
+++ b/build_thirdparty.sh
@@ -39,3 +39,14 @@ set -x
 
 # shellcheck disable=SC2086
 python3 "$YB_THIRDPARTY_DIR/python/yugabyte_db_thirdparty/yb_build_thirdparty_main.py" "$@"
+
+if [[ $* == *--snyk* ]]; then
+  if [[ $OSTYPE == linux* ]]; then
+    curl https://static.snyk.io/cli/latest/snyk-linux -o snyk
+  else
+    curl https://static.snyk.io/cli/latest/snyk-macos -o snyk
+  fi
+  chmod +x ./snyk
+  homedir="${BASH_SOURCE[0]%/*}"
+  ./snyk test "$homedir/src" --unmanaged
+fi
--- a/build_thirdparty.sh
+++ b/build_thirdparty.sh
@@ -40,6 +40,10 @@ set -x
 # shellcheck disable=SC2086
 python3 "$YB_THIRDPARTY_DIR/python/yugabyte_db_thirdparty/yb_build_thirdparty_main.py" "$@"
 
+rc=$?
+set +x
+
+
 if [[ $* == *--snyk* ]]; then
   if [[ $OSTYPE == linux* ]]; then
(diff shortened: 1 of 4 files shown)


build_thirdparty.sh
```diff
@@ -39,3 +39,32 @@ set -x
 
 # shellcheck disable=SC2086
 python3 "$YB_THIRDPARTY_DIR/python/yugabyte_db_thirdparty/yb_build_thirdparty_main.py" "$@"
+
+rc=$?
```
mbautin, line 43: The vast majority of our shell scripts run in the `set -euo pipefail` mode so that we don't have to check return codes after every commad. For this reason, in particular, grabbing the error code here is useless, because it cannot be different than zero. The script would have terminated in that case.

build_thirdparty.sh
```diff
@@ -39,3 +39,32 @@ set -x
...
+
+set +xue
```
mbautin, line 45: Let's not switch to the `set +ue` mode (allowing undefined identifiers, ignoring errors). This is a much more error prone way to write scripts.

build_thirdparty.sh
```diff
@@ -39,3 +39,32 @@ set -x
...
+if [[ -n "$SNYK_TOKEN" ]]; then
```
mbautin, line 46: `if [[ -n "${SNYK_TOKEN:-}" ]]` so it would work in `set -u` mode.

build_thirdparty.sh
```diff
@@ -39,3 +39,32 @@ set -x
...
+  fi
+  chmod +x ./snyk
+  homedir="${BASH_SOURCE[0]%/*}"
+
+  ./snyk auth "${SNYK_TOKEN}"
+  rc2=$?
+  
+  if [[ $rc2 -ne 0 ]]; then
+    exit $rc
+  fi
```
mbautin, line 61: This is exactly the kind of code that we don't need when running in the `set -e` mode. I think it is better to have exit-on-error by default.

build_thirdparty.sh
```diff
@@ -39,3 +39,32 @@ set -x
...
+set +xue
+if [[ -n "$SNYK_TOKEN" ]]; then
+  echo "Running Snyk Vulnerability scan"
+  if [[ $OSTYPE == linux* ]]; then
+    curl https://static.snyk.io/cli/latest/snyk-linux -o snyk
+  else
+    curl https://static.snyk.io/cli/latest/snyk-macos -o snyk
+  fi
+  chmod +x ./snyk
+  homedir="${BASH_SOURCE[0]%/*}"
```
mbautin, line 54: We already have a variable `${YB_THIRDPARTY_DIR}` that is exactly the same, if you take a look at the earlier part of this script.

build_thirdparty.sh
```diff
@@ -39,3 +39,32 @@ set -x
...
+  ./snyk auth "${SNYK_TOKEN}"
+  rc2=$?
+  
+  if [[ $rc2 -ne 0 ]]; then
+    exit $rc
+  fi
+
+  ./snyk monitor "$homedir/src" --unmanaged
+  rc=$?
+  rm -f ./snyk
```
mbautin, line 65: Not needed, could just leave it around.

build_thirdparty.sh
```diff
@@ -39,3 +39,32 @@ set -x
...
+fi
+
+set -xue
+
+exit $rc
```
mbautin, line 70: Not needed.

mbautin (changes requested):

PR #121, opened 2022-05-27
Update glog and gperftools, remove thin LTO and GCC 10 builds

mbautin:


mbautin:
Updating gperftools to fix a compilation issue with -std=c++20 specified on Linux aarch64.
Updating glog based on @es1024's work in https://github.com/yugabyte/yugabyte-db-thirdparty/pull/120.
( Comment from that PR: updating glog to 0.4.0-yb-2 to include yugabyte/glog#7 removing email functionality. Context: yugabyte/yugabyte-db#11930 )
Removing thin LTO build types because we are not using them yet. In theory they might be useful for development but we mostly use non-LTO builds for development and full LTO for production. Also remove GCC 10 builds as we only build with GCC 11 or later now.

svarnau (commented):
LGTM

PR #125, opened 2022-07-15
Modify library_checking to play nicely on NixOS

DAlperin:
Nixos does not follow conventional FHS meaning the assumptions about library paths can't be trusted. This patch fixes that. I'm working on packaging Yugabyte for Nixos, so this is my first patch towards that end. The next patch is an equally small change to the yugabyte db proper that I will file sometime soon.

Also in the interest of making less assumptions, this patch makes sure that the env is set correctly.

### DIFF
--- a/python/yugabyte_db_thirdparty/library_checking.py
+++ b/python/yugabyte_db_thirdparty/library_checking.py
@@ -93,8 +93,10 @@ def compile_re_list(re_list: List[str]) -> Any:
 
 
 def get_needed_libs(file_path: str) -> List[str]:
+    env = {"PATH": os.getenv("PATH")}
     return capture_all_output(
         ['patchelf', '--print-needed', file_path],
+        env=env,
         allowed_exit_codes={1},
         extra_msg_on_nonzero_exit_code="Warning: could not determine libraries directly "
                                        f"needed by {file_path}")
@@ -317,6 +319,9 @@ class LibTestLinux(LibTestBase):
             "^.* => /lib/",
             "^.* => /usr/lib/x86_64-linux-gnu/",
             "^.* => /opt/yb-build/brew/linuxbrew",
+            "^.* => /nix/store/",
+            "/bin/sh:",
+            "^.*/nix/store/",
             f"^.* => {re.escape(YB_THIRDPARTY_DIR)}"
         ]
 


mbautin:
Thanks for the contribution and sorry for the delay! Can you share some of the instructions of how to set up and configure NixOS to run the build?

mbautin:
Also what is FHS?

DAlperin:
@mbautin here is an example of how I built YugabyteDB on NixOS, the instructions are in comments at the top of each file: https://gist.github.com/DAlperin/b63cf1d383d7e4108d0241c69f12918a

FHS is the filesystem standard most systems use. For example,
/bin
/lib
/usr
/opt
/sys
etc...
NixOS does not conform to this standard, thus breaking software which assumes the FHS is present. For a better explanation, see here https://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard

mbautin:
@DAlperin thanks for the clarification. What is a good way to get a usable NixOS development environment in a Docker container? I tried `docker run. -it nixos/nix` but I am not sure if that gives me the environment that would allow to use the `.nix` files you provided.

python/yugabyte_db_thirdparty/library_checking.py
```diff
@@ -93,8 +93,10 @@ def compile_re_list(re_list: List[str]) -> Any:
 
 
 def get_needed_libs(file_path: str) -> List[str]:
+    env = {"PATH": os.getenv("PATH")}
     return capture_all_output(
         ['patchelf', '--print-needed', file_path],
+        env=env,
```
mbautin, line 99: Why do we need to remove all environment variables except PATH?

python/yugabyte_db_thirdparty/library_checking.py
```diff
@@ -317,6 +319,9 @@ def __init__(self) -> None:
             "^.* => /lib/",
             "^.* => /usr/lib/x86_64-linux-gnu/",
             "^.* => /opt/yb-build/brew/linuxbrew",
+            "^.* => /nix/store/",
+            "/bin/sh:",
```
mbautin, line 323: It would be nice to understand where this pattern comes from.

python/yugabyte_db_thirdparty/library_checking.py
DAlperin: We don't per se, but in my use case the PATH variable was not being picked up unless I explicitly included it, like so

python/yugabyte_db_thirdparty/library_checking.py
DAlperin: /nix/store is where everything in nixos lives including libraries, the `/bin/sh:` pattern is to exclude warnings generated by the tooling attempting to use /bin/sh which was doing something weird on nixos IIRC

DAlperin:
> What is a good way to get a usable NixOS development environment in a Docker container?

I don't know that you can run nixos itself under docker. But it might work with just the nix package manager as provided by the docker image? I'm not sure.

DAlperin:
@mbautin, just checking on this

PR #131, opened 2022-09-10
Fix CentOS 7 TSAN Clang 13 build

mbautin:
Needed to fix yugabyte/yugabyte-db#13615.

With Clang 13 and CentOS 7, the symbol __cxa_thread_atexit_impl in libcxxabi is referenced as a regular undefined symbol, not a weak symbol, leading to a runtime error. Explicitly set LIBCXXABI_HAS_CXA_THREAD_ATEXIT_IMPL as OFF in that case to avoid this problem. The resulting use of a weak symbol becomes consistent with what happens in non-sanitizer and ASAN builds, as well as on RHEL 8 based systems.

Also automatically save build log to a file in the ~/logs directory and create "latest" symlinks to it in build_thirdparty.sh.

### DIFF
--- a/build_thirdparty.sh
+++ b/build_thirdparty.sh
@@ -52,7 +52,7 @@ log_path=${log_dir}/${log_file_name}
 (
   cd "${log_dir}"
   for link_name in "${latest_log_links[@]}"; do
-    ln -sfT "${log_file_name}" "${link_name}"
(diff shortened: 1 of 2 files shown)


hari90:
Can you remove the tee commands from the Readme?

PR #142, opened 2022-09-28
SEC-267 Fix Snyk Static Code Analysis runs

bkms2018:
Since Github secrets are not available in PR runs from forked repositories, we need to scan code on a regular basis using a schedule.

### DIFF
--- a/.github/workflows/linux_build.sh
+++ b/.github/workflows/linux_build.sh
@@ -4,6 +4,7 @@ set -euo pipefail
 
 df -H .
 checkout_dir=$PWD
+YB_DOWNLOAD_EXTRACT_ONLY=${DOWNLOAD_EXTRACT_ONLY:-}
 echo "Building in directory: $checkout_dir"
 docker run -t \
   --cap-add=SYS_PTRACE \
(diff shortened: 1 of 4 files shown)


.github/workflows/linux_build.sh
```diff
@@ -4,13 +4,15 @@ set -euo pipefail
...
 checkout_dir=$PWD
+export YB_DOWNLOAD_EXTRACT_ONLY=${DOWNLOAD_EXTRACT_ONLY:-}
 echo "Building in directory: $checkout_dir"
 docker run -t \
   --cap-add=SYS_PTRACE \
   -e GITHUB_TOKEN \
   -e SNYK_TOKEN \
   -e YB_BUILD_THIRDPARTY_ARGS \
   -e YB_THIRDPARTY_ARCHIVE_NAME_SUFFIX \
+  -e YB_DOWNLOAD_EXTRACT_ONLY \
```
mbautin, line 15: I think this should be called "skip upload".

mbautin (approved):

.github/workflows/linux_build.sh
bkms2018: Updated the variable name

PR #148, opened 2022-10-22
Update to a new Linuxbrew archive repackaged with more locales

mbautin:


andrei-mart (approved):
LGTM

mbautin:
We will be removing Linuxbrew soon.

PR #156, opened 2023-01-11
DB-3937:Double Free issue with curl version

sureshdash2022-yb:
As part of Synk tool scanning Double Free issue for the curl version [0,curl-7.79.0)

Here is the detail.
Affected versions of this package are vulnerable to Double Free. When sending data to an MQTT server, libcurl <= 7.73.0 and 7.78.0 could in some circumstances erroneously keep a pointer to an already freed memory area and both use that again in a subsequent call to send data and also free it again.

Here is the Synk scanning report:-
https://security.snyk.io/vuln/SNYK-UNMANAGED-CURL-2317564

As part of the fix we will upgrade the curl version to 7.86


svarnau (approved):

sureshdash2022-yb:
Verified locally, so closing the PR

PR #157, opened 2023-01-16
DB-3971: Bump Kerberos version to 5.1-20.1

sureshdash2022-yb:
This PR upgrades the Kerberos version to 5.1-20.1


svarnau:
Tests indicate a new system lib is being pulled in: libzstd.so.1
Not sure if that should be added to allowed list or not.

dr0pdb:
@svarnau I think this PR is ready for review. I rebased it on latest master.

>Tests indicate a new system lib is being pulled in: libzstd.so.1

I couldn't find this in the logs. I think `libzstd` might have been getting pulled due to a different bug which was fixed in https://github.com/yugabyte/yugabyte-db-thirdparty/pull/162. Please let me know if that's not the case.

svarnau (approved):

PR #164, opened 2023-02-17
Update Abseil and TCMalloc to include new tcmalloc lifetime profiler.

SrivastavaAnubhav:
The TCMalloc lifetime profiler provides information on the lifetime of sampled allocations. It is necessary to get an equivalent of the "in_use_bytes" view of `yb-prof`, which shows objects that were allocated but not deallocated during profiling.

The lifetime profiler requires some additional abseil dependencies. The new abseil version exposes all dependencies so we don't have to keep updating thirdparty every time we need a new function.


python/build_definitions/tcmalloc.py
```diff
@@ -41,7 +41,6 @@ def build(self, builder: BuilderInterface) -> None:
                 dep=self, src_file="tcmalloc_static.a", dest_file="libgoogletcmalloc.a",
                 src_folder="tcmalloc", is_shared=False)
 
-        # Copy headers.
-        builder.log_output(log_prefix, ['mkdir', '-p', builder.prefix_include + '/tcmalloc'])
-        builder.log_output(log_prefix, ['cp'] + glob.glob('./tcmalloc/*.h') +
-                                       [builder.prefix_include + '/tcmalloc'])
+        # Copy headers, keeping the folder structure. https://stackoverflow.com/a/29457076.
+        builder.log_output(log_prefix, ["rsync", "-a", "--include=*.h", "-f",
+                                        "hide,! */", "./tcmalloc", builder.prefix_include])
```
mbautin, line 46: I think the best discussion for this is at https://stackoverflow.com/questions/9622883/recursive-copy-of-a-specific-file-type-maintaining-the-file-structure-in-unix-li/9626253#9626253

Also I don't get how "hide" works.

mbautin (approved):

python/build_definitions/tcmalloc.py
SrivastavaAnubhav: Will update that comment to point to that discussion. There's a description of what `-f 'hide,! */'` does on a comment there too. In short, `-f 'hide,! */'` is a filter that excludes all files (`*/` matches all directories and `!` negates it to match all files), and the `--include=*.h` adds back only the headers.

PR #165, opened 2023-02-24
Add HdrHistogram_c library

es1024:
Adding HdrHistogram/HdrHistogram_c to build to replace our partial port of the Java version (HdrHistogram/HdrHistogram), context: https://github.com/yugabyte/yugabyte-db/issues/16087.


python/build_definitions/hdrhistogram.py
```diff
@@ -0,0 +1,35 @@
...
+from yugabyte_db_thirdparty.build_definition_helpers import *  # noqa
+
+
+class HdrHistogramDependency(Dependency):
+    def __init__(self) -> None:
+        super(HdrHistogramDependency, self).__init__(
+            'hdrhistogram',
+            '0.11.6-yb-1',
+            'https://github.com/yugabyte/HdrHistogram_c/archive/hdrhistogram-{0}.tar.gz',
+            BUILD_GROUP_COMMON)
```
mbautin, line 24: Can we put it in the UNINSTRUMENTED group so that it gets ASAN/TSAN instrumentation? (Those groups are named in a somewhat misleading way.)

mbautin (approved):

PR #175, opened 2023-03-30
Upgrade Boost, OpenSSL, support GCC 12 and Clang 16

mbautin:
Component upgrades:
- Upgrade OpenSSL to 1.1.1t.
- Upgrade Boost to 1.81.0 to stop relying on std::unary_function and std::binary_function, which are deprecated and are not supported by LLVM 16's libc++.

Other necessary changes for compiler upgrades:
- Add a patch to remove the unnecessary un-definition of `__STRICT_ANSI__` from ufile.cpp in icu4c (needed for compilation to succeed under GCC 12).
- libedit build needs an additional include directory to find ncurses headers when building with Clang 16.
- When looking for the `libclang_rt.ubsan_minimal[-<arch>].so` shared library, consider multiple different possible directories, as the library directory layout may differ between different versions of Clang.

Refactoring and minor improvements:
- Simplify the get_install_prefix_with_qualifier function to remove an unused `qualifier` parameter.
- Support `--add-checksums` and `--add-checksum` options interchangeably.
- Remove a file specific to LLVM 7 (we dropped support for LLVM 7 a long time ago).


python/build_definitions/bison.py
```diff
@@ -25,5 +25,18 @@ def __init__(self) -> None:
             license='GPL-3.0')
         self.copy_sources = True
 
+    def get_additional_compiler_flags(self, builder: BuilderInterface) -> List[str]:
+        llvm_major_version = builder.compiler_choice.get_llvm_major_version()
+        linux_llvm16_or_later = (
```
hari90, line 30: nit: can combine into a single if statement

python/build_definitions/bison.py
```diff
@@ -25,5 +25,18 @@ def __init__(self) -> None:
...
+            is_linux() and llvm_major_version is not None and llvm_major_version >= 16)
+        flags = []
```
hari90, line 32: nit: can combine into a single if statement

python/build_definitions/bison.py
```diff
@@ -25,5 +25,18 @@ def __init__(self) -> None:
...
+        if linux_llvm16_or_later:
```
hari90, line 33: nit: can combine into a single if statement

python/yugabyte_db_thirdparty/clang_util.py
```diff
@@ -39,26 +43,61 @@ def get_clang_library_dirs(clang_executable_path: str) -> List[str]:
...
-        for subdir_name in ['linux', arch_specific_subdir_name]:
+        for subdir_name in subdir_names:
             candidate_dir = os.path.join(library_dir, 'lib', subdir_name)
-            if os.path.isdir(candidate_dir):
-                return candidate_dir
+            if os.path.isdir(candidate_dir) and (
+                    look_for_file is None or
+                    os.path.exists(os.path.join(candidate_dir, look_for_file))):
+                if all_dirs:
+                    found_dirs.append(candidate_dir)
```
hari90, line 80: `found_dirs` is not needed. Since the else anyway returns, we can skip the if case and use `candidate_dirs` instead of `found_dirs`

SrivastavaAnubhav (approved):
Maybe mention llvm7 is being removed in the commit.

python/build_definitions/cassandra_cpp_driver.py
```diff
@@ -69,4 +69,9 @@ def get_additional_cxx_flags(self, builder: 'BuilderInterface') -> List[str]:
             # Needed to avoid this error:
             # https://gist.githubusercontent.com/mbautin/d1ce54c995f9e535ab214a12945d2e7b/raw
             extra_cxx_flags.append('-Wno-error=free-nonheap-object')
+            if gcc_major_version >= 12:
```
hari90, line 72: Do we need to check version? Do the old compilers complain when passed an unknown flags? If not, we can always add the flags

Same for `llvm_major_version >= 16`

hari90 (approved):

python/build_definitions/bison.py
mbautin: Done.

python/yugabyte_db_thirdparty/clang_util.py
mbautin: found_dirs are all directories satisfying the condition, and candidate_dirs are all the directories considered whether or not they satisfy the condition, for logging the error.

python/build_definitions/cassandra_cpp_driver.py
mbautin: Yes, old compilers complain if passed an unknown flag. And I would like to keep the behavior for old compilers the same. But we can remove support for compilers older than Clang 14 now.

mbautin: Or, since this is GCC, I guess, older than GCC 11? Anyway, I prefer to keep the logic for old compiler versions the same to avoid having to re-test if the build still works for old compilers.

PR #182, opened 2023-04-26
Fix how we locate Clang runtime library directory

mbautin:
Also add a branch.txt file and logic in build_and_release.sh to use it, as specified in https://github.com/yugabyte/yugabyte-db-thirdparty#setting-up-a-new-release-version-branch


svarnau (approved):

HarshDaryani896 (commented):
Is the `branch.txt` file added? Could not find it.

PR #191, opened 2023-05-18
Compile opentelemetery libraries for distributed tracing of YSQL queries

abhinab-yb:
[Opentelemetry](https://opentelemetry.io/) is a CNCF project that defines a standard for distributed tracing. The project also provides API and SDKs to instrument code in various languages including C++. It's license is Apache 2.0 and permission has been given to use it in Yugabyte.

This commit compiles the latest (v1.9.0) version of opentelemetry-cpp. This version uses a protobuf feature experimental_allow_proto3_optional that is not supported by the protobuf version (v3.5.1) used in Yugabyte. Therefore a couple of patches to remove the usage of the feature has also been added. Another patch to fix undefined symbols error on macOS was added.


yb-thirdparty-common.sh
```diff
@@ -75,7 +75,6 @@ ensure_correct_mac_architecture() {
   fi
   if [[ $YB_TARGET_ARCH != "x86_64" && $YB_TARGET_ARCH != "arm64" ]]; then
     fatal "Invalid value of YB_TARGET_ARCH on macOS (expected x86_64 or arm64): $YB_TARGET_ARCH"
-    exit 1
```
svarnau, line 78: Why was this removed?

abhinab-yb: From the commit message of @vrajat `Remove exit 1 as fatal already exits. Caught by shellcheck`, it seems like `exit 1` won't be called. Should this be reverted?

svarnau: Oh, I see. That is a call to fatal, not just log message. I just wanted to make sure it was intentional.

svarnau (approved):

vrajat (approved):

PR #206, opened 2023-07-20
Draft: Revert "Do not rely on patch marker file if local_archive is set for a dependency."

dr0pdb:
Reverts yugabyte/yugabyte-db-thirdparty#199

Just trying to revert the commit to see if this is causing the macOS build to break in https://github.com/yugabyte/yugabyte-db-thirdparty/pull/205.


dr0pdb:
The CI in macOS still failed, so this diff is not the issue.

PR #228, opened 2023-09-19
Upgrade OpenSSL to 1.1.1w; add GCC 13 support

mbautin:
Also automatically set the path to Abseil source tree when building TCMalloc, so there is no need to keep updating the WORKSPACE file in the TCMalloc repo every time Abseil is updated.


python/build_definitions/tcmalloc.py
```diff
@@ -18,17 +18,41 @@
...
         super(TCMallocDependency, self).__init__(
             name='tcmalloc',
-            version='e116a66-yb-4',
+            version='e116a66-yb-5',
             url_pattern='https://github.com/yugabyte/tcmalloc/archive/{0}.tar.gz',
             build_group=BuildGroup.POTENTIALLY_INSTRUMENTED)
         self.copy_sources = True
         self.bazel_project_subdir_name = 'com_google_tcmalloc'
 
+    def update_workspace_file(self) -> None:
```
SrivastavaAnubhav, line 32: Please add a comment describing what this does

SrivastavaAnubhav (approved):

PR #230, opened 2023-09-22
[BACKPORT 2.14] Add abseil

SrivastavaAnubhav:
This commit backports the following commits from master (excluding the tcmalloc parts):
 - Add Abseil to thirdparty build (33ef9a18cfdc87b33d67f65084ee9a9942c703e4)
 - File Abseil file naming (5791d35ac9474250ea2fc86c2cd90f8c686c4da5)
 - Copy .inc files when copying abseil headers too (f77b09b3635b8f35ef254d42648930f82c712696)

The Abseil version being introduced is `abseil-0064d9d-yb-3`.


python/yugabyte_db_thirdparty/builder.py
```diff
@@ -860,12 +914,10 @@ def init_linux_clang_flags(self, dep: Dependency) -> None:
...
                 "libc++")
-            self.ld_flags += ['-lc++', '-lc++abi']
-
-            self.cxx_flags = [
-                '-stdlib=libc++',
-                '-nostdinc++'
-            ] + self.cxx_flags
+            self.ld_flags += ['-stdlib=libc++', '-lc++', '-lc++abi']
+            # TODO(asrivastava): We might not need libc++ in cxxflags but removing it causes certain
+            # builds to fail.
```
mbautin, line 919: Why don't we need it? We need to tell the complier to use libc++.

mbautin (approved):

python/yugabyte_db_thirdparty/builder.py
SrivastavaAnubhav: I can see that it is being ignored during compilation of some C++ files. For example:
```
[abseil (uninstrumented)] INFO: From Compiling absl/time/internal/cctz/src/time_zone_format.cc:
[abseil (uninstrumented)] clang: warning: argument unused during compilation: '-stdlib=libc++' [-Wunused-command-line-argument]
```

Is this something we should investigate?

PR #264, opened 2024-05-03
Enforce the use of C++20 for building all C++ dependencies

mbautin:
We were previously building Boost with C++14, and potentially also building other dependencies with C++ standard versions other than C++20, which is currently used across the main YugabyteDB codebase. This is a suspected root cause of failures to catch exceptions thrown from Boost.

Relevant issues: https://github.com/yugabyte/yugabyte-db/issues/22191 and https://github.com/yugabyte/yugabyte-db/issues/13064. For the older issue https://github.com/yugabyte/yugabyte-db/issues/13064 a custom workaround ( e.g. see the commit https://github.com/yugabyte/llvm-project/commit/8efd9000cde37c1999ce886fbd0fc33712cb1d2f ) was implemented in libc++abi. It is possible that it would be unnecessary with this fix.

Also removing most of the Clang 16 builds (only keeping the one on Ubuntu 20.04).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -55,22 +55,6 @@ jobs:
               --devtoolset=11
               --expected-major-compiler-version=11
 
-          # Clang 16
-          - name: centos7-x86_64-clang16
-            os: ubuntu-20.04  # Ubuntu 20.04 is for the top-level VM only. We use Docker in it.
-            docker_image: yugabyteci/yb_build_infra_centos7_x86_64:v2023-10-25T03_33_02
-            build_thirdparty_args: >-
-              --toolchain=llvm16
-              --expected-major-compiler-version=16
-
-          - name: centos7-x86_64-clang16-full-lto
-            os: ubuntu-20.04  # Ubuntu 20.04 is for the top-level VM only. We use Docker in it.
-            docker_image: yugabyteci/yb_build_infra_centos7_x86_64:v2023-10-25T03_33_02
-            build_thirdparty_args: >-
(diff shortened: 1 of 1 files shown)


es1024 (approved):

python/yugabyte_db_thirdparty/compiler_choice.py
```diff
@@ -30,6 +30,7 @@
 )
 from packaging.version import parse as parse_version
 
+
```
OlegLoginov, line 33: Let's clean-up - remove this new line. (Minor optional comment.)

python/yugabyte_db_thirdparty/compiler_flag_util.py
```diff
@@ -0,0 +1,62 @@
...
+def remove_incorrect_cxx_standard_flags(cmd_args: List[str]) -> List[str]:
+    """
+    Returns a new list of compiler arguments that is obtained from the given list by removing all
+    arguments that specify an incorrect version of the C++ standard. All of the dependencies we
+    are building must use the same version of the C++ standard.
+    """
+    return [
+        arg for arg in cmd_args
+        if (not is_cxx_standard_flag(arg) or
+            is_correct_cxx_standard_version(get_cxx_standard_version_from_flag(arg)))
```
OlegLoginov, line 61: Is it a correct `if` check?
Keep `arg`: if (Not standard flag)  OR (correct std version)

OlegLoginov (commented):
LGTM. Just 2 minor notes&questions.

PR #265, opened 2024-05-06
[DEVOPS-3069][BACKPORT 2.20] Add LLVM17 build types, plus enable amazonlinux2

svarnau:
To enable 2.20 build support of amazonlinux2, enable clang17, since clang16 has a bug preventing compilation on AL2.


svarnau:
Cancelled workflow, because I neglected to add workflow for amzn2

svarnau:
> @svarnau can we remove Clang 16 because we will be exclusively using Clang 17 to build 2.20 from this point on?

I think we only need clang17 for AL2-aarch64, so at this point I think it would be best to leave the other flavors using clang16.

svarnau:
Original commit: f9a1b05070b763b774c37683c5987b7e71af3540

PR #266, opened 2024-05-07
Add DiskANN

mbautin:
Adding the DiskANN dependency (approximate nearest neighbor search). Currently only supporting x86_64 Linux, no ASAN/TSAN. Build support for macOS, aarch64 and ASAN/TSAN will be added later. 

Intel MKL and OpenMP libraries are required by DiskANN on x86_64, provided by the Intel oneAPI package. Creating a way to identify the needed subset of Intel oneAPI, package it as an archive, and download that archive during the build. We are not building Intel libraries from scratch but using pre-built binary versions of these libraries.

In addition to Intel MKL and OpenMP, we are also adding another library required by DIskANN, libaio, which is a Linux library for asynchronous I/O. We are adding it to the thirdparty build system in the usual way.

This commit also includes a lot of refactoring of the thirdparty build system. In particular, the compiler wrapper script now has a feature that allows to collect all header file paths used during the build process. This is used to collect the subset of Intel oneAPI headers required by DiskANN.

Another independent improvement on the CI/CD side is the added support of asterisk wildcards for the "CI build types" commit message keyword.

### DIFF
--- a/.github/workflows/github_actions_build.sh
+++ b/.github/workflows/github_actions_build.sh
@@ -56,8 +56,9 @@ if [[ $GIT_HEAD_COMMIT_MESSAGE == *"$CI_BUILD_TYPES_KEYWORD"* ]]; then
     if [[ $build_type_pattern == *linux* ]]; then
       echo >&2 "Expanding pattern '$build_type_pattern' to match all Linux distributions."
       build_type_patterns_array+=(
-        "${build_type_pattern//linux/centos}"
         "${build_type_pattern//linux/almalinux}"
+        "${build_type_pattern//linux/amazonlinux}"
+        "${build_type_pattern//linux/centos}"
         "${build_type_pattern//linux/ubuntu}"
       )
     fi


.github/workflows/github_actions_build.sh
```diff
@@ -6,20 +6,99 @@ readonly CI_BUILD_TYPES_KEYWORD="CI build types:"
...
+    build_type_pattern=${build_type_pattern#"${build_type_pattern%%[![:space:]]*}"}
+    build_type_pattern=${build_type_pattern%"${build_type_pattern##*[![:space:]]}"}
+
+    build_type_patterns_array+=( "$build_type_pattern" )
+    if [[ $build_type_pattern == *linux* ]]; then
+      echo >&2 "Expanding pattern '$build_type_pattern' to match all Linux distributions."
+      build_type_patterns_array+=(
+        "${build_type_pattern//linux/centos}"
+        "${build_type_pattern//linux/almalinux}"
+        "${build_type_pattern//linux/ubuntu}"
```
svarnau, line 61: Could add amazonlinux, but not required at this point.

svarnau (approved):
LGTM

PR #269, opened 2024-06-16
[2.18] Compile libev with EV_USE_EVENTFD=0

es1024:


python/build_definitions/libev.py
```diff
@@ -25,7 +25,8 @@ def __init__(self) -> None:
         self.copy_sources = True
 
     def build(self, builder: BuilderInterface) -> None:
-        builder.build_with_configure(dep=self, extra_args=['--with-pic'])
+        builder.build_with_configure(
+            dep=self, extra_args=['CPPFLAGS=-DEV_USE_EVENTFD=0', '--with-pic'])
```
jharveysmith, line 29: This should just be the option I think

es1024: This runs `./configure CPPFLAGS=-DEV_USE_EVENTFD=0 --with-pic` which sets `CPPFLAGS=-DEV_USE_EVENTFD=0` in the Makefile, and results in `-DEV_USE_EVENTFD=0` passed in to the compiler itself to set `EV_USE_EVENTFD=0`

Huqicheng (approved):
LGTM

PR #271, opened 2024-06-17
Fix Mac build errors

hari90:
Fix bison, gettext, and openldap build definitions for mac.

### DIFF
--- a/python/build_definitions/bison.py
+++ b/python/build_definitions/bison.py
@@ -28,9 +28,10 @@ class BisonDependency(Dependency):
     def get_additional_compiler_flags(self, builder: BuilderInterface) -> List[str]:
         llvm_major_version = builder.compiler_choice.get_llvm_major_version()
         flags = []
-        if (is_linux() and
-                llvm_major_version is not None
-                and llvm_major_version >= 16):
+        linux_llvm16_or_later = (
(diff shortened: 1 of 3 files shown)


python/build_definitions/bison.py
```diff
@@ -28,9 +28,10 @@ def __init__(self) -> None:
     def get_additional_compiler_flags(self, builder: BuilderInterface) -> List[str]:
         llvm_major_version = builder.compiler_choice.get_llvm_major_version()
         flags = []
-        if (is_linux() and
-                llvm_major_version is not None
-                and llvm_major_version >= 16):
+        linux_llvm16_or_later = (
+            is_linux() and llvm_major_version is not None and llvm_major_version >= 16)
```
SrivastavaAnubhav, line 32: Why do you need to change this if you're already setting `Wno-error=incompatible-function-pointer-types` in gettext.py?

SrivastavaAnubhav (changes requested):

python/build_definitions/bison.py
hari90: We set it per build. Its not persisted across different build definitions

SrivastavaAnubhav (approved):

PR #278, opened 2024-08-19
[#21963] YSQL: Add build definition for AWS clockbound FFI.

pao214:
https://github.com/aws/clock-bound lets us retrieve clock accuracies. This is accurate as long as the reference clock is accurate. This information helps us improve transaction performance.

Compiling clock-bound-ffi using rust generates a C library for use in YugabyteDB. The source code is licensed Apache 2.0.

As a part of this change, build clock-bound-ffi using rust. We package only the header file and the shared library (not the static library for llvm version safety).


pao214:
Superseded by https://github.com/yugabyte/yugabyte-db-thirdparty/pull/280
Undesirable to create new branches in the main repo.

PR #282, opened 2024-09-18
[#21963] Add macos-arm64 to ci.yml

pao214:


svarnau (approved):
Great! The did not previously support arm64, so we had to do an out-of-band build.

PR #286, opened 2024-10-14
[skip ci] Setup 2024.1 branch

es1024:


svarnau:
Is there a reason we need a 2024.1 branch, or can the newest thirdparty used on master also be used for 2024.1 (and 2024.2)?

PR #290, opened 2024-10-22
[2.20] Remove CircleCI, move to GitHub Actions for aarch64

es1024:
c78311cee06fb8dfd2fedf5a5b1b68a19eaf066a moved aarch64 builds from CircleCI (which no longer works) to GitHub Actions. This does the same for 2.20 branch.

Also removed CentOS 7 builds since it's EOL.


.github/workflows/ci.yml
```diff
@@ -83,6 +100,23 @@ jobs:
               --expected-major-compiler-version=17
               --lto=full
 
+          - name: centos7-aarch64-clang17
```
svarnau, line 103: The centos7 entries should be removed. It is EOL and we no longer support it on any release. The amzn2 OS should be used as the lowest version glibc to be compatible with.

svarnau (approved):
Please remove centos7 entries.

PR #304, opened 2025-02-20
Upgrade to C++23

es1024:
Build thirdparty using C++23, so that the main repo can be updated to using C++23 as well. A compile error was fixed in tcmalloc, and some build code was changed to build mac with --std=c++2b instead of --std=c++23, since the latter is not supported on Apple Clang. CMake normally does this automatically, but we override this flag for thirdparty.


SrivastavaAnubhav (approved):
If / when we get rid of apple clang, we should switch back to just 23

PR #305, opened 2025-02-21
[WIP] Build OS X + clang 19

es1024:


es1024:
> What are the assertion errors for which you're disabling assertions for libunwind?
```
libunwind: malformed __unwind_info at 0x10072B318 bad second level page
```
from https://github.com/yugabyte/llvm-project/blob/llvmorg-21.1.1-yb-2/libunwind/src/UnwindCursor.hpp#L1923

SrivastavaAnubhav:
Is it the same as this? https://github.com/dask/distributed/issues/2608. Is there an libunwind issue you can link?

es1024:
Going to go ahead and merge this since the uncatchable exception issue turns out to be a problem on Linux as well and isn't mac-specific; will fix that issue separately.

PR #307, opened 2025-04-01
Add BID library

hari90:
Add BID (Binary Integer Decimal) from the  intelmathlib for non-aarch64 build types.
This is required for the DocumentDB extension.

- Updated the `download_dependency` function to handle patches with CRLF line  endings.
- Added 2 patch files, one for each type of line endings.
- Removed `$(FO)$@` from makefile since this causes `-c xxx.o` to get generated.
- Removed `CC_TYPE` from makefile since we pass the required arguments directly.
- Added `#include <signal.h>` and `#include <stdlib.h>` in appropriate places.


patches/bid.patch
```diff
@@ -0,0 +1,70 @@
...
+-_CPP := $(CC_NAME) $(call CcNameSelect,-EP,-EP,-E -P,-EP)
+-_CC  := $(CC_NAME) $(call CcNameSelect,, -nologo,, -nologo)
+-
+-ifeq ($(origin CC),default)
+-    CC := $(_CC)
+-endif
+-ifeq ($(origin CPP),default)
+-    CPP := $(_CPP)
+-endif
+-
```
SrivastavaAnubhav, line 35: Why did you need to remove these?

patches/bid-crlf.patch
SrivastavaAnubhav, line 1: Why is this called crlf? It removes the unnecessary .o argument, right? Please mention that somewhere in the PR

python/build_definitions/bid.py
```diff
@@ -0,0 +1,60 @@
...
+
+    def get_additional_compiler_flags(self, builder: BuilderInterface) -> List[str]:
+        flags = ['-Wno-error=missing-braces']
+        return flags
+
+    def build(self, builder: BuilderInterface) -> None:
+        with PushDir("LIBRARY"):
+            builder.build_with_make(
+                dep=self,
+                extra_make_args=['_CFLAGS_OPT=-fPIC',
```
SrivastavaAnubhav, line 39: Does just -fPIC not work?

python/yugabyte_db_thirdparty/download_manager.py
```diff
@@ -375,9 +375,14 @@ def download_dependency(
             with PushDir(src_path):
                 for patch in dep.patches:
                     log("Applying patch: %s", patch)
-                    process = subprocess.Popen(['patch', '-p{}'.format(dep.patch_strip)],
-                                               stdin=subprocess.PIPE)
-                    with open(os.path.join(YB_THIRDPARTY_DIR, 'patches', patch), 'rt') as inp:
+                    process = subprocess.Popen(
+                        ['patch',
+                         '--binary',
```
SrivastavaAnubhav, line 380: Why do we need --binary?

patches/bid.patch
hari90: Because we use clang11, clang13, or compiler_wrapper.py as the CC

patches/bid-crlf.patch
hari90: CRLF is the windows line endings, which is why I need 2 patch files. The other one is linux LR line endings.

python/yugabyte_db_thirdparty/download_manager.py
hari90: CRLF

python/build_definitions/bid.py
hari90: This is how DocumentDB built it. Do you have a strong preference on this?

hari90: Just `-fPIC` does not work

`make: PIC: No such file or directory`

SrivastavaAnubhav (approved):

PR #308, opened 2025-04-05
Instrument BSON and BID for Sanitizer builds

hari90:
This is required to use the libraries in ASAN and TSAN builds


SrivastavaAnubhav:
This isn't required, you could also put it in `common` instead of `uninstrumented`, but this is better anyways.

SrivastavaAnubhav (approved):

PR #311, opened 2025-04-24
Use CMake 3 on OS X

es1024:
CMake 4 removes compatibility for CMake older than 3.5, but we have thirdparty dependencies that require that compatibility. On OS X, we end up installing CMake 4 indirectly when we invoke brew, and there is no easy way to pin CMake version there. Instead, this change changes some old build code we had to avoid 3.19.1 by installing CMake manually, to also avoid 4.x.


kai-franz (approved):
LGTM!

PR #323, opened 2025-09-16
Add GCC 15 builds via GccInstaller

es1024:
Adds GCC 15 builds using builds from GccInstaller, via `--toolchain=gcc15`. This works similarly to LlvmInstaller clang builds: we download pre-built releases of GCC+binutils from https://github.com/yugabyte/build-gcc releases to build everything. The old way of building with system GCC (`--compiler-{prefix,family,suffix}` or `--devtoolset`) remain unmodified and available, for cases like testing building on different architectures.

Required changes to build with GCC 15:
- missing include in DiskANN
- silencing some new warnings in a couple of dependencies
- adding `-std=gnu17` for some dependencies with C code, since GCC 15 uses C23 by default and C23 is not fully backwards compatible (in particular: `void f()` is now `void f(void)` not `void f(...)`).


python/yugabyte_db_thirdparty/builder.py
```diff
@@ -214,6 +214,8 @@ def determine_compiler_family_and_prefix(self) -> Tuple[str, Optional[str]]:
             self.toolchain.write_url_and_path_files()
             if self.args.toolchain.startswith('llvm'):
                 compiler_family = 'clang'
+            else:
```
hari90: Safer to check starts with gcc.
And if it's neither then let's throw an error

python/yugabyte_db_thirdparty/toolchain.py
```diff
@@ -24,21 +24,22 @@
...
-
-
-def get_llvm_url(tag: str) -> str:
-    return 'https://github.com/yugabyte/build-clang/releases/download/%s/yb-llvm-%s.tar.gz' % (
-            tag, tag)
+INSTALLER_VERSION_FROM_ARCHIVE_NAME_RE = re.compile(
+        rf'^yb-(?:llvm|gcc)-v(.*)-[0-9]+-[0-9a-f]+-.*')
 
 
 MIN_LLVM_VERSION = 14
```
hari90: Do we still want the old min versions? Why not bump it up?

hari90 (approved):

PR #351, opened 2026-04-20
Use libunwind from toolchain

es1024:
For llvm-installer clang builds, we currently build libunwind with the llvm_libunwind dependency. This libunwind was slightly different from the toolchain libunwind we link to, leading to cases where we could not catch exceptions.

This picks up changes to our LLVM build's libunwind (that mirror the changes that thirdparty libunwind had) and changes llvm_libunwind to simply copy the libunwind library from the toolchain.

### DIFF
--- a/python/build_definitions/llvm_libunwind.py
+++ b/python/build_definitions/llvm_libunwind.py
@@ -13,55 +13,45 @@
 
 import os
 
-from build_definitions.llvm_part import LlvmPartDependencyBase
 from yugabyte_db_thirdparty.build_definition_helpers import *  # noqa
 
 
-class LlvmLibUnwindDependency(LlvmPartDependencyBase):
+class LlvmLibUnwindDependency(Dependency):
     def __init__(self, version: str) -> None:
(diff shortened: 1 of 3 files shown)


python/build_definitions/llvm_libunwind.py
```diff
@@ -13,55 +13,41 @@
...
-        builder.build_with_cmake(
-            self,
-            extra_cmake_args=extra_cmake_args,
-            src_subdir_name=src_subdir_name)
-
-        # TODO: do not use this "standalone" build of libunwind -- it is deprecated.
-        # https://github.com/yugabyte/yugabyte-db/issues/11962
-        # Maybe we won't need to do this manual copying of headers if we use a supported approach.
-        src_include_path = os.path.join(source_path, src_subdir_name, 'include')
+        # This is only None for non-llvm-installer builds, and this dependency is
```
SrivastavaAnubhav, line 29: Could you add a top-level comment about what the code below does?

SrivastavaAnubhav (approved):

PR #352, opened 2026-04-20
[skip ci] Set up 2025.1 branch

es1024:


svarnau (commented):
Why do we need a 2025.1 branch, if there is no real code change here?  It is in maintenance only another 3 months. Is there some dependency change coming?

PR #358, opened 2026-06-04
Add DuckDB as a thirdparty dependency

HarshDaryani896:
Builds DuckDB v1.4.3 once as a thirdparty artifact (`libduckdb_bundle.a` plus DuckDB's internal header tree) instead of recompiling it from source on every yugabyte-db build. Consumed by the pg_duckdb PostgreSQL extension.

- Version pin matches upstream pg_duckdb's DuckDB submodule pointer.
- Compiled-in extensions (json, icu, httpfs) stay in sync with pg_duckdb's `pg_duckdb_extensions.cmake`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


python/build_definitions/duckdb.py
```diff
@@ -0,0 +1,145 @@
...
+    GIT_TAG 9c7d34977b10346d0b4cbbde5df807d1dab0b2bf
+)
+"""
+
+
+class DuckDBDependency(Dependency):
+    def __init__(self) -> None:
+        super(DuckDBDependency, self).__init__(
+            name='duckdb',
+            version=DUCKDB_VERSION,
```
es1024: Put the version inline here as it's what all the other definitions do.

python/build_definitions/duckdb.py
```diff
@@ -0,0 +1,145 @@
...
+from yugabyte_db_thirdparty.build_definition_helpers import *  # noqa
+from yugabyte_db_thirdparty.env_helpers import EnvVarContext
+
+
+# DuckDB is consumed by the pg_duckdb PostgreSQL extension in the yugabyte-db repo. Rather than
+# embedding DuckDB's ~14k source files in yugabyte-db (as a submodule or subtree) and recompiling it
+# on every fresh build, we build it once here and publish libduckdb_bundle.a plus DuckDB's internal
+# header tree as a thirdparty artifact. pg_duckdb relies on DuckDB's private headers (not just the
+# public amalgamation), so the full src/include tree is shipped.
+#
```
es1024: No need to mention why it's a thirdparty dependency vs submodule/subtree here, this should be the default choice unless there's a compelling reason otherwise.

python/build_definitions/duckdb.py
```diff
@@ -0,0 +1,145 @@
...
+            # command terminator (DuckDB's Makefile uses the same single-quoting idiom elsewhere).
+            "-DCMAKE_PREFIX_PATH='" + common_dir + ';' + builder.prefix + "'",
+        ])
+
+        build_env = {
+            # Stamps the DuckDB version into the binary without requiring a git checkout.
+            'OVERRIDE_GIT_DESCRIBE': 'v' + DUCKDB_VERSION,
+            'GEN': 'ninja',
+            'CMAKE_VARS': cmake_vars,
+            'DISABLE_SANITIZER': '1',
```
es1024, line 127: Why?

python/build_definitions/duckdb.py
```diff
@@ -0,0 +1,145 @@
...
+            'DISABLE_ASSERTIONS': '0',
```
es1024: probably 1, we do -NDEBUG for everything in thirdparty

python/build_definitions/duckdb.py
```diff
@@ -0,0 +1,145 @@
...
+# yugabyte-db repo. httpfs is fetched from its own git repo during the DuckDB CMake configure step
+# (it provides S3/HTTP read support that pg_duckdb depends on).
+PG_DUCKDB_EXTENSIONS_CMAKE = """\
+duckdb_extension_load(json)
+duckdb_extension_load(icu)
+duckdb_extension_load(httpfs
+    GIT_URL https://github.com/duckdb/duckdb-httpfs
+    GIT_TAG 9c7d34977b10346d0b4cbbde5df807d1dab0b2bf
+)
+"""
```
es1024, line 43: Ideally this is something we commit to a fork instead.

python/build_definitions/duckdb.py
```diff
@@ -0,0 +1,145 @@
...
+        }
+
+        # Build DuckDB's objects without LTO, dropping the -flto flag the thirdparty build injects
+        # into CFLAGS/CXXFLAGS for LTO build types. DuckDB ships as a prebuilt static archive
+        # (libduckdb_bundle.a) that pg_duckdb links in the yugabyte-db repo, so LTO of these objects
+        # buys nothing here and is actively harmful: LTO bitcode is tied to an exact LLVM version,
+        # so an archive built with it would fail to link if yugabyte-db's compiler ever diverges
+        # from thirdparty's. Full-LTO clang+libc++ also instantiates a unique_ptr<CachedFileHandle>
+        # destructor against an incomplete type in the pinned httpfs sources, which only the LTO
+        # instantiation path rejects. Emitting regular objects sidesteps both issues.
```
es1024: I don't think this is a good reason to disable LTO.

We always build thirdparty with same compiler as yugabyte-db -- the build code always downloads the archive for a specific (os, compiler, architecture, lto type) tuple.

LTO also is relevant whether or not we static or dynamically link. In fact, it's _more_ relevant when we statically link, since the compiler has a chance to optimize across not just compilation units within duckdb, but with pg_duckdb as well. That's not possible if you don't compile with -flto.

Re: unique_ptr<CachedFileHandle> destructor, this is something that'd need a code patch (it's usually because you did ~Class() = default; in the header that doesn't have access to the definition, solved by moving it into the source file). It's technically not specific to LTO though, and compiler version bumps often surface this issue as well.

python/build_definitions/duckdb.py
```diff
@@ -0,0 +1,145 @@
...
+
+        self.install_artifacts(builder, src_path)
+
+    def install_artifacts(self, builder: BuilderInterface, src_path: str) -> None:
+        # The single static bundle pg_duckdb links against.
+        bundle_lib = os.path.join(src_path, 'build', 'release', 'libduckdb_bundle.a')
+        if not os.path.exists(bundle_lib):
+            raise IOError("Expected DuckDB bundle library not found: %s" % bundle_lib)
+        mkdir_p(builder.prefix_lib)
+        shutil.copy(bundle_lib, os.path.join(builder.prefix_lib, 'libduckdb_bundle.a'))
```
es1024, line 149: Would have expected there be something similar to make install/cmake --install that would do this copying. But this is fine I guess.

python/yugabyte_db_thirdparty/dependency_selection.py
```diff
@@ -90,6 +90,9 @@ def get_final_dependency_module_names(compiler_choice: CompilerChoice) -> List[s
         'eigen',
         'cnpy',
         'libxml2',
+        # Consumed by the pg_duckdb PostgreSQL extension. Built once here instead of recompiling
+        # DuckDB from source on every yugabyte-db build.
+        'duckdb',
```
es1024: Comment unnecessary.

python/build_definitions/duckdb.py
```diff
@@ -0,0 +1,145 @@
...
+            mkdir_p(os.path.join(src_path, 'build', 'release', 'vcpkg_installed'))
+            with EnvVarContext(build_env):
+                # `bundle-library` builds the `release` target and packs every static lib
+                # (libduckdb_static.a, third_party/*, extensions, vcpkg) into libduckdb_bundle.a.
+                builder.log_output(log_prefix, ['make', 'bundle-library'])
+
+        self.install_artifacts(builder, src_path)
+
+    def install_artifacts(self, builder: BuilderInterface, src_path: str) -> None:
+        # The single static bundle pg_duckdb links against.
```
es1024: Comment unnecessary.

es1024 (changes requested):

python/build_definitions/duckdb.py
HarshDaryani896: ~~DuckDB defaults `ENABLE_SANITIZER`/`ENABLE_UBSAN` to `TRUE` (its `CMakeLists.txt`), so without `DISABLE_SANITIZER=1` it self-instruments with `-fsanitize=address,undefined` even in release. Our own sanitizer flags still come through `CXXFLAGS` for sanitizer build types, so this only avoids double-applying them.~~

python/build_definitions/duckdb.py
HarshDaryani896: Agreed on keeping LTO — reverting the disable-LTO change.

Root cause: in httpfs's `http_state.hpp`, `CachedFile::GetHandle()` is defined inline while `CachedFileHandle` is only forward-declared there, so the returned `unique_ptr<CachedFileHandle>`'s destructor instantiates against an incomplete type — exactly the case you described. clang+libc++ rejects it under full-LTO. Fix: move the body into `http_state.cpp` where `CachedFileHandle` is complete (verified locally — the two failing TUs compile clean under `-flto=full -std=c++23 -stdlib=libc++` with the patch).

Since httpfs is git-fetched during configure (not in the DuckDB tarball), rather than a separate fork I'd like to use DuckDB's own `APPLY_PATCHES` mechanism: add `APPLY_PATCHES` to the httpfs `duckdb_extension_load` and stage the patch into `<duckdb>/.github/patches/extensions/httpfs/` at build time, so DuckDB's `apply_extension_patches.py` applies it after the fetch. Keeps the fix self-contained as a reviewable `.patch`, with no external fork to maintain.

---
_automated · Claude Code (Opus 4.8)_

python/build_definitions/duckdb.py
es1024: Please add a comment about this then, since it's pretty misleading.

python/build_definitions/duckdb.py
es1024: Makes sense. Please add a comment with some of this explanation (httpfs not part of tarball) why patches are used instead of fork for this.

HarshDaryani896: Comment added

patches/duckdb-icu-undef-reserved-category-macros.patch
```diff
@@ -0,0 +1,44 @@
...
+ #define _Sk     UPROPS_FLAG(U_MODIFIER_SYMBOL)
+ #define _So     UPROPS_FLAG(U_OTHER_SYMBOL)
+ #define _Pi     UPROPS_FLAG(U_INITIAL_PUNCTUATION)
++
++/*
++ * YugabyteDB: undefine the single-underscore general-category convenience macros above.
++ * They use reserved identifiers (e.g. _Sc) that collide with standard library headers --
++ * upstream ICU already disables _Lo/_Pe/_Pc/_Pf for the same reason (MSVC headers). With
++ * clang's libc++, _Sc collides with a template parameter in <__algorithm/radix_sort.h>, and
++ * DuckDB's unity build leaks the macro into a later TU that includes <algorithm>. These
```
es1024, line 13: Something like this really does look like it should be in a fork instead of patch.

But also, can you verify that an LTO build with duckdb works (that it builds + runs fine)? I'm not sure if there are any issues here since we actually build ICU (and also snappy) ourselves, and it looks like duckdb will bring in its own version.

HarshDaryani896: Verified an LTO build end-to-end — it builds **and** runs fine; the bundled ICU/snappy don't conflict with thirdparty's.

**Setup:** built this thirdparty branch with `--toolchain=llvm21 --lto=full`, then built yugabyte-db `--clang21 --thin-lto` against it and ran pg_duckdb in a live cluster.

**Build/link:**
- `yb-server` (thin-LTO) links and `pg_duckdb.so` builds against the prebuilt `libduckdb_bundle.a` (pg_duckdb's Makefile errors out if the bundle is missing — it does not rebuild the submodule).
- `ldd pg_duckdb.so` shows **no** separate `libicu*`/`libsnappy` — DuckDB's ICU and snappy are statically inside the bundle, so there's no duplicate *dynamic* dependency alongside thirdparty's `icu4c`/`snappy`.

**Runtime (live YB cluster, pg_duckdb preloaded):**
- `CREATE EXTENSION pg_duckdb` + backend startup load the `.so` (with its statically-bundled ICU/snappy) with no symbol conflict.
- snappy: `COPY ... (FORMAT PARQUET)` then `read_parquet(...)` round-trips; confirmed the file's codec is `SNAPPY` (so both bundled-snappy encode and decode ran).
- `read_csv` (local) works.
- httpfs (the patched extension) reads a remote file over HTTPS fine.

So the "DuckDB brings its own ICU/snappy" worry doesn't materialize — they're statically bundled and isolated within pg_duckdb.so.


---
_automated · Claude Code (Opus 4.8)_

es1024 (approved):
Accepting to unblock but please move the patch and pg_duckdb_extensions.cmake into a yb fork as followup cleanup.

python/build_definitions/duckdb.py
HarshDaryani896: @es1024 Have raised this PR to correct the comment: https://github.com/yugabyte/yugabyte-db-thirdparty/pull/365
Also created a GHI https://github.com/yugabyte/yugabyte-db-thirdparty/issues/364 to instrument duckdb build later.

PR #359, opened 2026-06-09
Bump curl to 8.20.0 and ncurses to 6.6

shishir2001-yb:
Upgrades two third-party dependencies to their latest upstream releases:

- curl: 8.19.0 -> 8.20.0
- ncurses: 6.4 -> 6.6

ncurses 6.5+ changed two defaults, so two configure flags are added to preserve existing behavior:

- `--disable-widec`: ncurses now builds the wide-char libncursesw by default. libedit links against the narrow libncurses, and without this flag it falls back to the system libncurses/libtinfo, failing the library dependency check.
- `--disable-db-install`: ncurses 6.6 fails `make install` if /usr/share/terminfo is not writable (macOS CI, non-root builds). We use the system terminfo database at runtime and never ship the installed one, so skip installing it.


proff4eg (approved):
LGTM

jharveysmith (approved):

PR #361, opened 2026-06-25
Add include-what-you-use for clang

ttyusupov:
Added include-what-you-use for clang: https://github.com/include-what-you-use/include-what-you-use


python/build_definitions/iwyu.py
```diff
@@ -0,0 +1,99 @@
...
+
+class IncludeWhatYouUseDependency(Dependency):
+    """
+    include-what-you-use (IWYU): a Clang-based tool that analyzes #include usage. It is shipped as a
+    build-time host tool (not linked into YugabyteDB), so it is built once in the
+    CXX_UNINSTRUMENTED group, like patchelf.
+
+    IWYU pins to Clang's internal (unstable) APIs, so its release must match the toolchain's Clang
+    major version: IWYU 0.25 targets Clang/LLVM 21. For that reason it is only selected for Clang 21
+    builds (see the version gate in dependency_selection.py); bump both together on a Clang upgrade.
```
es1024: Can we make version depend on compiler version instead, e.g. have version=None and something like:
```
    VERSIONS_MAP = { 19: '...', 21: '0.25', ... }

    def dependency_version(self, builder: BuildInterface) -> Optional[str]:
        if builder.compiler_choice.is_llvm_installer_clang():
            return VERSIONS_MAP[builder.compiler_choice.get_llvm_major_version()]
        raise RuntimeError('only llvm-installer clang supported')
```
(and then have builder check .dependency_version() if .version is None)

We actually do in some cases (like today: on 2026.1/2025.2) use multiple clang versions from the same thirdparty commit. So would rather not indirectly hardcode the toolchain version for a thirdparty commit.

python/yugabyte_db_thirdparty/dependency_selection.py
```diff
@@ -101,4 +101,10 @@ def get_final_dependency_module_names(compiler_choice: CompilerChoice) -> List[s
     if not is_building_for_aarch64():
         dep_names.append('bid')
 
+    # include-what-you-use pins to Clang's internal APIs, so it can only be built with a matching
+    # Clang and not with GCC (no Clang libraries to link). IWYU 0.25 targets Clang 21; keep this
+    # version gate in sync with the version in build_definitions/iwyu.py.
+    if compiler_choice.is_clang() and compiler_choice.get_llvm_major_version() == 21:
```
es1024: is_llvm_installer_clang()

es1024 (changes requested):

python/build_definitions/iwyu.py
```diff
@@ -0,0 +1,121 @@
...
+        # it patches is identical across the IWYU releases we build, so one patch covers all.
+        self.patches = ['iwyu-add-targetparser-link-component.patch']
+
+    def dependency_version(self, builder: BuilderInterface) -> str:
+        compiler_choice = builder.compiler_choice
+        if not compiler_choice.is_llvm_installer_clang():
+            raise RuntimeError(
+                'include-what-you-use can only be built with an LLVM-installer Clang toolchain')
+        major: Optional[int] = compiler_choice.get_llvm_major_version()
+        version = IWYU_VERSION_FOR_CLANG_MAJOR.get(major) if major is not None else None
```
es1024: nit: `IWYU_VERSION_FOR_CLANG_MAJOR.get(major)` returns None if not found (and None is not found anyways)

python/yugabyte_db_thirdparty/dependency_selection.py
```diff
@@ -101,4 +101,10 @@ def get_final_dependency_module_names(compiler_choice: CompilerChoice) -> List[s
     if not is_building_for_aarch64():
         dep_names.append('bid')
 
+    # include-what-you-use pins to Clang's internal APIs, so it can only be built with an
+    # LLVM-installer Clang (not GCC, which has no Clang libraries to link). The matching IWYU
+    # release is chosen per Clang major version in build_definitions/iwyu.py.
+    if compiler_choice.is_llvm_installer_clang():
```
es1024: Sorry for the back and forth on this: I've actually just merged a cleanup change to replace all is_llvm_installer_clang() with is_clang() by dropping support for Apple Clang. So this will need to be reverted back to is_clang().

es1024 (approved):
Couple of small changes but otherwise looks good.

